PR #1, opened 2018-02-02
License, gitignore, git secret

melissachang:


hooks/README.md
```diff
@@ -0,0 +1,3 @@
+# git secrets
+* If you have not done so already please [install git-secrets](https://github.com/awslabs/git-secrets#installing-git-secrets)
+* To ensure git secrets is run please copy or link this directory to .git/hooks/ locally.
```
calbach: nit: A copy-pastable command-line would be more useful here.

melissachang: Done

PR #17, opened 2018-03-08
Setup API server (including Swagger) and nginx proxy

melissachang:
Similar to job manager, use nginx proxy to get around CORS for local development.

Tested:
* View localhost:4400 works in Chrome. UI server calls API server and I see response in Console.
* Above also works with <hostname>:4400
* Swagger UI works

@bfcrampton can you review since CH and Alan are out? @calbach -- feel free to review and I can do follow-up PRs to address your comments.


README.md
```diff
@@ -1,4 +1,41 @@
...
+TODO(melissachang): Add Javascript definitions for React.
+
+1. If you do not already have the jar, you can download it here:
+  ```
+  # Linux
+  wget http://central.maven.org/maven2/io/swagger/swagger-codegen-cli/2.2.3/swagger-codegen-cli-2.2.3.jar -O swagger-codegen-cli.jar
+  # macOS
+  brew install swagger-codegen
+  ```
+1. Clear out existing generated models:
```
bfcrampton: These aren't rendering correctly in github unfortunately. I think there's a bug that when numbered lists contain code blocks they don't increment automatically. Mind just switch to `1,2,3...`?

README.md
```diff
@@ -1,4 +1,41 @@
 # Data explorer
 
-## One-time setup
+## Architecture overview
+
+For local development, nginx is used to get around CORS.
+
+![Architecture overview](https://i.imgur.com/tKv8FEL.png)
```
bfcrampton: Nice, we should add this to the job manager as well

api/api.yaml
```diff
@@ -0,0 +1,17 @@
...
+  version: '0.0.1'
+paths:
+  /facets:
+    get:
+      description: Returns facets.
+      responses:
+        200:
+          description: Facets.
+          schema:
+            type: string
```
bfcrampton: nit: If you add a `tags` here it will swagger codegen will auto generate different controllers per tag instead of putting it all in default_controller. Ex: https://github.com/DataBiosphere/job-manager/blob/master/api/jobs.yaml#L26

api/api.yaml
```diff
@@ -0,0 +1,17 @@
...
+    url: 'http://opensource.org/licenses/BSD-3-Clause'
+  version: '0.0.1'
+paths:
+  /facets:
+    get:
+      description: Returns facets.
+      responses:
+        200:
+          description: Facets.
+          schema:
```
bfcrampton: I think this should return an array of facet strings instead of a single string?

ui/.dockerignore
```diff
@@ -0,0 +1 @@
+node_modules
```
bfcrampton: May also want to add `swagger-codegen-cli.jar`

README.md
bfcrampton: Realized you probably took part of this from the JM readme which has the same issue: https://github.com/DataBiosphere/job-manager#updating-the-api-using-swagger-codegen

api/api.yaml
melissachang: Will handle in future PRs. I'll create a Facets object which has list of Facet, etc.

ui/.dockerignore
melissachang: I changed README instructions to put swagger-codegen-cli.jar in ~, since it doesn't make to have a separate jar for each repo.

melissachang:
Addressed comments

PR #19, opened 2018-03-09
Add Elasticsearch and Kibana to docker-compose.

melissachang:
Load test data into Elasticsearch.

Next PR will have API server read from Elasticsearch.


test/indexer.py
```diff
@@ -0,0 +1,42 @@
...
+from elasticsearch import Elasticsearch
+from elasticsearch import helpers
+
+
+INDEX_NAME = 'test'
+
+
+def init_elasticsearch():
+  # Wait for Elasticsearch to come up.
+  time.sleep(10)
```
bfcrampton: nit: Is there a way you can ping the server every say second to see if its up yet?

docker-compose.yml
```diff
@@ -14,13 +24,15 @@ services:
...
-    links:
-      - ui
-      - api
+      - 9200:9200
+  kibana:
+    image: docker.elastic.co/kibana/kibana-oss:6.2.2
+    ports:
+      - 5601:5601
+  # For convenience, load test data into Elasticsearch.
+  index_test_data:
```
bfcrampton: How do you plan to swap out indexers here?

test/indexer.py
melissachang: Added comment. The right way is to use wait_for_status, but I [had trouble getting that to work](https://discuss.elastic.co/t/cant-get-python-client-wait-for-status-to-work/123163). I think sleep is good enough for now. FYI the right number is somewhere between 5 and 10 seconds, so we're not losing that many seconds.

docker-compose.yml
melissachang: I will add instructions later. It will look like:
- Before running Data explorer, write index to some directory.
- Somehow point Elasticsearch to that directory.
- Somehow point to dataset config directory which has dataset name.

FYI each index will have a different name and an Elasticsearch instance can have many indices. So there could be a test index, baseline index, onduo index, etc. I figure I can write the "test" index every time since it's fast. It's ok if it's not used.

test/indexer.py
bfcrampton: That's unfortunate. This is definitely fine for this purpose, but if you're looking for a workaround I had a (kind of hacky) workaround on the old data explorer repo: 

https://github.com/bvprivate/data-explorer-service/blob/9cc7ffc373222d419ea196d8d2e897e064d99c63/data_index/test/in_memory_elastic.py#L21-L41

bfcrampton: (despite the bad name, that is actually a real ES not in-memory)

PR #21, opened 2018-03-14
Have API server query Elasticsearch and return facets

melissachang:
Unlike Boardwalk, I'm using the FacetedSearch library (https://elasticsearch-dsl.readthedocs.io/en/latest/faceted_search.html). This simplifies client code. We'll see if this library works out.

Tested: http://localhost:8390/api/facets works


test/dataset.json
```diff
@@ -0,0 +1,7 @@
+{
+  // Elasticsearch index will be given this name.
```
alahwa: The JSON language does not support comments, can we remove them from here and document them in the README if necessary? That would eliminate the need to parse them in the indexer, which is pretty fragile as it doesn't support the full range of in-line comments.

api/data_explorer/swagger/swagger.yaml
```diff
@@ -17,8 +17,32 @@ paths:
...
         200:
-          description: "Facets."
+          description: "Success"
           schema:
-            type: "string"
+            $ref: "#/definitions/FacetsResponse"
       x-swagger-router-controller: "data_explorer.controllers.facets_controller"
-definitions: {}
+definitions:
+  FacetsResponse:
```
alahwa: This definition is a bit redundant. I'm not an expert on API definition but consider removing this and then changing the /facets 200 response to something like:

```
description: "Success"
schema:
  type: array
  items:
    $ref: "#/definitions/Facet"
```

api/data_explorer/swagger/swagger.yaml
```diff
@@ -17,8 +17,32 @@ paths:
...
+    type: "array"
+    description: "Results from a faceted search."
+    items:
+      $ref: "#/definitions/Facet"
+  Facet:
```
alahwa: On the topic of naming, since the concept of a Facet will be pretty integral across our code base here, I'd like to be convinced that this is the best choice before we invest too heavily to change this. Facet fits syntactically with faceted search which is nice, are there any other reasons why we chose this naming? If the goal of the data explorer is solely to provide faceted search on datasets I think this is a good name, otherwise we might opt for something more widely recognized.

Alternatively, I like the name 'Field' as this is an instantly recognizable concept to anybody who has worked with tables and encapsulates the same idea. We could also use 'Feature' which is more familiar to ML but also conveys the same idea. Thoughts?

README.md
```diff
@@ -16,26 +35,27 @@ update the server implementations:
...
       -o api \
       -DsupportPython2=true,packageName=data_explorer
     ```
-1. Update the server implementations to resolve any broken dependencies on old API definitions or implement additional functionality to match the new specs.
+* Update the server implementations to resolve any broken dependencies on old API definitions or implement additional functionality to match the new specs.
 
 ### One-time setup
-[Set up git secrets.](https://github.com/DataBiosphere/data-explorer/tree/master/hooks)
+
+* Install `swagger-codegen-cli.jar`.
```
alahwa: Add this to the .dockerignore

api/data_explorer/controllers/facets_controller.py
```diff
@@ -1,15 +1,24 @@
...
 
 
 def facets_get():
     """
-    facets_get
     Returns facets.
 
-    :rtype: str
+    Returns:
+        List of Facet.
```
alahwa: nit: Facets

api/data_explorer/controllers/facets_controller.py
```diff
@@ -1,15 +1,24 @@
...
     """
-    return 'do some magic!'
+    search = DatasetFacetedSearch()
+    response = search.execute()
+    facets = []
+    for facet_name, values in response.facets.to_dict().iteritems():
+        facet_values = []
+        for value_name, count, _ in values:
```
alahwa: What's the third value here expected by unpacking ', _'? I don't see a third field in the API spec.

api/data_explorer/dataset_faceted_search.py
```diff
@@ -0,0 +1,33 @@
...
+    # (https://plus.google.com/+DouglasCrockfordEsq/posts/RK8qyGVaGSr). Remove
+    # comments before parsing.
+    with open('/app/dataset.json') as f:
+        dataset = json.loads('\n'.join([row for row in f.readlines() if len(row.split('//')) == 1]))
+        return dataset['name']
+
+
+class DatasetFacetedSearch(FacetedSearch):
+    index = _get_index_name()
+    using = Elasticsearch('elasticsearch:9200')
```
alahwa: Seems like something we shouldn't be hard coding, as the elasticsearch server won't always be local. Can we extract this?

api/data_explorer/swagger/swagger.yaml
melissachang: I'm thinking about what api calls we need to the support. I'm pretty sure we'll need a text box that will autocomplete on field name.

A dataset may have hundreds of fields. 10 of those fields might be facets.

The UI calls /facets to specifically request the facet fields.

To support autocomplete, there could be a /fields call that searches over all hundreds of fields.

I think this should be called Facet, because the UI specifically wants facets to display in a facet UI. (And I think the React component should also be Facet.)

test/dataset.json
melissachang: I think it's much easier -- for both Data explorer user and Data explorer developer -- if comments are directly in the JSON, vs having to switch back and forth between README and JSON. I don't think not supporting all possible comment formats is a problem. If people are going to add a comment, most likely they'll copy existing format.

alahwa: IIRC the indexer requires the name and primary_key to be present: since we've already documented what should be contained in this file elsewhere, adding the additional documentation here isn't strictly necessary. In fact, if I were a new user, it might be convenient for me to be able to have this information presented in a README so that I don't have to go digging through the config parsing code to figure out how my dataset should be structured.

If we think comments are going to be necessary for clarity here, I would do it this way:
- The JSON spec does not support comments, and this config will be un-parseable as JSON by any other tooling out there. Because of this, we should change the extension of this file from .json to .txt or something similar. We want to avoid bad surprises when my SomeTool.asJson('dataset.json') returns a failure.
- We should clearly document the JSON-based configuration language extended to support comments that is required by the dataset.txt file. Different languages have lots of different ways of representing comments, and we shouldn't assume that everybody will pick up that ^// in-line comments are the only supported comment format here.

As far as I can tell, Elasticsearch documents follow the JSON spec and do not have comments, in which case I would still suggest just dropping the comments, letting JSON be JSON, and documenting what a dataset.json should look like. Since this is based on the Elasticsearch document, it shouldn't be totally unfamiliar to anybody setting this up.

melissachang: To summarize offline discussion:

- I think it's much easier to have README say "cp -r config directory. Fill out config files according to comments.", rather than "Copy/paste this template for file 1. Here are 10 descriptions for 10 fields in file 1. Copy/paste this template for file 2....". Also, I think it's better for maintainability to have comments inline. Note the config files will become much more complicated over time.
- To support a variety of comment styles, I will use jsmin. This is how the JSON creator recommends handling JSON comments.

PR #24, opened 2018-03-16
Fix typo in ui/Dockerfile

melissachang:


ui/a/b/c.txt
```diff
@@ -0,0 +1 @@
+sfdsf
```
alahwa: What's this file for?

melissachang: Oops, removed

PR #26, opened 2018-03-21
 Read facets from facet_fields.csv

melissachang:
For test index, for numeric fields, change Elasticsearch index type (string -> long).
Read config files once during lifetime of Flask server, instead of on every request.


README.md
```diff
@@ -12,12 +12,13 @@ To use a different dataset:
...
-* In `api`, create a symlink named `config` that points to a config directory similar to
-[this one](https://github.com/DataBiosphere/data-explorer-indexers/tree/8f22de4ad7750c60ab4d73d6608cb154436f68af/bigquery/config/template).
-Specifically:
-
+* In `api`, create a symlink named `config` that points to the `config`
+directory from the above step. See [example](https://github.com/DataBiosphere/data-explorer-indexers/blob/master/bigquery/config/platinum_genomes/facet_fields.csv)
+here. Specifically:
   * There must be a file named `dataset.json` that has a `name` field. This
 determines the name of the Elasticsearch index.
+  * There must be a file named `facet_fields.csv` with the `readable_field_name`
```
alahwa: Since I believe we use these field names as keys, maybe mention that each should be unique? This is a breaking requirement on the UI and (probably?) the backend as well.

api/data_explorer/__main__.py
```diff
@@ -78,5 +81,17 @@ def split_env_flag(name):
 app.app.json_encoder = JSONEncoder
 app.add_api('swagger.yaml', base_path=args.path_prefix)
 
+@app.app.before_first_request
+def init():
+    # Read config files. Just do this once; don't need to read files on every
```
alahwa: IMO this is self-documenting between placement in init() and the annotation.

api/data_explorer/dataset_faceted_search.py
```diff
@@ -1,40 +1,78 @@
...
-    # TODO(melissachang): Get facets from facet_fields.csv.
-    facets = {
-        'Gender': TermsFacet(field='Gender.keyword'),
-        'Region': TermsFacet(field='Region.keyword'),
-    }
+
+    def __init__(self):
+        """Initializes DatasetFacetedSearch object.
+
+        Contains initialization that should be done once over the lifetime of
```
alahwa: Same as my comment above, people reading the Python here are probably familiar with `__init__(self)`.

api/data_explorer/dataset_faceted_search.py
```diff
@@ -1,40 +1,78 @@
...
+def get_index_name():
+    """Gets index name from dataset.json."""
+    with open('/app/config/dataset.json') as f:
         # Remove comments using jsmin, as recommended by JSON creator
         # (https://plus.google.com/+DouglasCrockfordEsq/posts/RK8qyGVaGSr).
         dataset = json.loads(jsmin.jsmin(f.read()))
         return dataset['name']
 
 
+def get_facets():
```
alahwa: I'm not too familiar with how we're structuring this area of the code base, is this function intended to replace the one in facets_controller? If so, can we either delete that or yank the get_facets logic out of here and extract it into the facets_controller, passing in this instance to prevent duplicate instantiations of this class?

api/data_explorer/controllers/facets_controller.py
```diff
@@ -18,9 +15,6 @@ def facets_get():
         List of Facets.
     """
     search = DatasetFacetedSearch()
-    search.using = Elasticsearch(current_app.config['ELASTICSEARCH_URL'])
```
alahwa: I noticed this PR breaks the UI. Are there references to this function that need to be updated to the facets_get in dataset_faceted_search?

test/indexer.py
```diff
@@ -1,11 +1,15 @@
 """For convenience, load test data into Elasticsearch index "test".
 
+Input must be JSON, not CSV. Unlike JSON, CSV values don't have types, so
```
alahwa: Beyond the scope of this test set, how do we expect most indexable to come in? As JSON exported from bigquery?

api/data_explorer/dataset_faceted_search.py
melissachang: Actually this comment is wrong, I moved the once-per-server initialization to to __main__ init(). Deleted comment.

test/indexer.py
melissachang: No, the Elasticsearch index will have been created before `docker-compose up`. Added a comment to the top of this file ("For real datasets,..."). I also updated the [README.md](https://github.com/DataBiosphere/data-explorer/blob/master/README.md); I was missing a couple steps. Do things make sense now?

PR #27, opened 2018-03-21
Implement static UI

alahwa:
screen shot 2018-03-21 at 1 58 01 am

- Cleanup function scoped var > block scoped let
- Delete unnecessary skeleton code
- Re: use of material, after reviewing the Clarity docs, it looks like library support is only available for Angular. Startup cost of embedding external stylesheets into each component and modifying for functionality and customization for the sake of using Clarity is too high given that we don't currently have real wireframes, let alone mockups. In the meantime, we can import and use material-ui parts out of the box to iterate fast and have a UI that works and looks pretty good, and not be too sad if we end up switching off of it.


ui/src/components/Header.css
```diff
@@ -0,0 +1,39 @@
...
+
+.logo {
+    width: 90%;
+    display: block;
+    margin-left: auto;
+    margin-right: auto;
+    margin-bottom: 15px;
+}
+
+.countBox {
```
melissachang: totalCountBox, totalCountBoxText, totalCount

ui/src/components/Header.css
```diff
@@ -0,0 +1,39 @@
+.header {
+    width: 45%;
+    margin-top: 15px;
+    vertical-align: center;
+}
+
+.logo {
```
melissachang: datasetLogo

ui/src/index.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
 import './index.css';
-import App from './App';
+import App from './containers/App';
```
melissachang: Why is App in containers? Seems like most React apps put in either components/ or src/.

melissachang:
Also, http://www.material-ui.com/#/ says "Material-UI v1 is coming". Up to you, but I prefer using the v1 version; might as well switch now instead of later. Might just be a matter of changing "material-ui" to "material-ui@next" in package.json.

ui/src/components/FacetsGrid.js
```diff
@@ -0,0 +1,26 @@
+import './FacetsGrid.css';
+
+import React from 'react';
+import FacetCard from "./FacetCard";
+import {GridList, GridTile} from "material-ui/GridList";
+
+function FacetsGrid(props) {
+    const facetResponse = props.facetResponse;
```
melissachang: facetResponse -> facetsResponse

Change everywhere

ui/src/index.js
alahwa: I'm dividing app structure here into Presentational/Container components as referenced [here](https://daveceddia.com/react-project-structure/). I think this is a good design for our needs, which are going to be a pretty even split between pure UI and custom logic. App.js is a logical container for stuff, and would go under containers.

How we should structure the app is open for discussion, although it's pretty trivial to change in the future and the app is small enough for now that it doesn't really matter.

ui/src/components/Header.css
alahwa: Done

ui/src/components/Header.css
alahwa: Done

ui/src/components/FacetsGrid.js

alahwa:
Also re: v1, looks like there are going to be more breaking changes introduced between the current beta to pre-release to official release. At the moment, v1 supports fewer features than the current 0.x, has an unknown release timeline, is in beta and unstable, is largely undocumented, and has less community adoption for troubleshooting. Given that the main benefit of v1 is improved performance over 0.x, I would suggest sticking with the stable version unless we have a convincing reason to update.

ui/src/index.js
melissachang: I did some investigation. I think the containers vs components distinction is mainly useful for Redux.

Out of the 12 sample apps at https://reactjs.org/community/examples.html, 3 use containers/. All 3 use redux. Out of the 9 that don't, 1 uses redux.

I find containers vs components confusing. A Facet is a "logical container for stuff"; why isn't that a container?

https://daveceddia.com/react-project-structure/ says "Containers are the stateful ones, and the ones that make the API calls.". But then that requires someone to know which React components make API calls. Component X could make an API call today, but if we rearranged the UI, X could not make an API call tomorrow.

Something about Redux forces people to make the distinction between "components-that-make-API-calls" and "component-don't-make-API-calls". Since we're not using Redux, I'd rather not make people think about that.

As a side note -- while things like this are easy to change in the future, I prefer to have the (short) discussion and change now, rather than after code is submitted.

alahwa: Okay, I moved `App.js` up to the `src/` directory and all of our components into `component/`. As we scale up, we'll organize components by feature hierarchy.

PR #29, opened 2018-03-22
Add API for filtering results (API not implemented yet)

melissachang:
This only adds API to api.yaml, it does not implement the API in backend or frontend. After this is in, work can begin on backend or frontend.

Tested:
* localhost:8390/api/facets?filter=Gender=female works
* React UI still works


api/api.yaml
```diff
@@ -10,6 +10,32 @@ paths:
   /facets:
     get:
       description: Returns facets.
+      parameters:
+        - name: filter
+          # Ideally this query parameter would use an object. However, OpenAPI
```
alahwa: This is all context behind this PR which can go into the PR description/comments instead of cluttering the API definition itself. Additionally, could you file an issue for updating to 3.0 and then changing this to an object?

ui/src/App.js
```diff
@@ -23,6 +23,9 @@ class App extends Component {
     var apiClient = new ApiClient();
     apiClient.basePath = '/api'
     var api = new FacetsApi(apiClient)
+    var opts = {
```
alahwa: I'm assuming the changes to this file were for testing purposes? If so, can we revert?

ui/src/api/README.md
```diff
@@ -72,14 +72,18 @@ var DataExplorerService = require('data_explorer_service');
```
alahwa: It seems strange to have a bunch of code to execute in the README. Was this autogenerated?

api/api.yaml
melissachang: I prefer to have Issues for short-term things (next month or so); the [Issues tab](https://github.com/DataBiosphere/data-explorer/issues) is getting kind of hard to read. For things that might never happen, or could happen in next year or two, I prefer TODO. WDYT?

I guess an alternative is to use labels and label every Issue as short-term or long-term, but it's simpler just to only file issues for short-term.

ui/src/App.js
melissachang: The change to line 37 was necessary. Without that change, UI doesn't work. Facets response is not printed in console.

ui/src/api/README.md
melissachang: Yes. It was in fact very helpful for figuring out how fix above UI problem.

api/api.yaml
alahwa: If you dislike github issues we could reconsider our issue tracking service. Regardless, I don't think paragraphs of context belong in the code base.

I generally follow the Google style guide for TODOs. That is, use them for temporary or short term code and always assign them to either a person or a bug. In this case, if you wanted to make this a TODO I would file the github issue and document it here as `TODO (#XX): Change to object on swagger update to 3.0`

ui/src/App.js
alahwa: Then can we just keep line 37 as`api.facetsGet([], callback)`? This function is the initial population of Facets, which should return all of them.

api/api.yaml
melissachang: Done. In retrospect I agree with you. I left one small comment because it's strange to have a string like "Gender=female"; normally things would be broken up into objects.

ui/src/App.js
melissachang: Done. Just be aware that ui/src/api/README.md has sample code.

PR #33, opened 2018-03-23
Minor tweaks to get Baseline demo working

melissachang:
* Remove facet card height, since some cards have many facets.
* Docker COPY doesn't support symlinks, so change instructions to copy files into config/.
* Change histogram value name to range. "10" -> "10 - 19".


api/data_explorer/controllers/facets_controller.py
```diff
@@ -18,8 +20,20 @@ def facets_get():
...
         for name, count, _ in values:
-            facet_values.append(FacetValue(name=name, count=count))
+            es_facet = current_app.config['ELASTICSEARCH_FACETS'][facet_name]
+            if isinstance(es_facet, HistogramFacet):
+              # For histograms, Elasticsearch returns:
+              #   name 10: count 15     (There are 15 people aged 10-19)
+              #   name 20: count 33     (There are 33 people aged 20-29)
+              # Convert "10" -> "10-19".
+              range_str = '%d-%d' % (name,
+                  name + es_facet._params['interval'] - 1)
```
alahwa: The assumption here being that numeric ranges are always in discreet whole numbers. Is this always the case? i.e. if my intervals are 0 - 0.1, 0.1 - 0.2, we wouldn't want to apply -1.

api/data_explorer/controllers/facets_controller.py
```diff
@@ -18,8 +20,20 @@ def facets_get():
     response = search.execute()
     facets = []
     for facet_name, values in response.facets.to_dict().iteritems():
+        # isinstance(current_app.config['ELASTICSEARCH_FACETS']['Age'], HistogramFacet)
```
alahwa: Remove commented code

api/data_explorer/dataset_faceted_search.py
```diff
@@ -48,7 +48,7 @@ def get_facets():
         else:
             # Assume numeric type.
             # TODO: Handle other types.
-            facets[field_name] = HistogramFacet(field=field_name, interval=7)
+            facets[field_name] = HistogramFacet(field=field_name, interval=10)
```
alahwa: Since this is a quick hack, could we add a TODO here to more intelligently assign bucket intervals depending on the range of the data that we're looking at? I'm not sure exactly what that would look like, maybe it would go in the config.

api/data_explorer/controllers/facets_controller.py
melissachang: Done

api/data_explorer/controllers/facets_controller.py
melissachang: Makes sense. Let's address this when we come across this.

api/data_explorer/dataset_faceted_search.py
melissachang: Added TODO. API server should figure it out on the fly. I don't want it go in configs, it would make configs more complicated. If people really wanted to specify custom ranges, we perhaps make it an optional config, though I would prefer not to.

For certain fields, we should hardcode ranges. For example, for age, "65-over" is common. Or for LDL cholesterol, there are standard ranges.

Issue #7, opened 2018-03-06
select facets and dynamically update results


dionloy:

melissachang:
Fixed by many PRs. Last one #47.

PR #35, opened 2018-03-27
Interactive UI for querying facets

alahwa:
The GridList maintains the state for all of its cards, and triggers an API call for get facets with the list of query params every time a checkbox is clicked. Note that until the API supports faceted search, the returned value is always the full list of facets.

### DIFF
--- a/ui/src/components/FacetsGrid.js
+++ b/ui/src/components/FacetsGrid.js
@@ -66,7 +66,6 @@ class FacetsGrid extends Component {
                 this.searchFilters.set(facetName, this.removeFacet(this.searchFilters.get(facetName), facetValue));
             }
         }
-        console.log(this.serialize(this.searchFilters));
         this.api.facetsGet({filter: this.serialize(this.searchFilters)}, this.callback);
     }
 
@@ -82,12 +81,14 @@ class FacetsGrid extends Component {
 
     serialize(searchFilters) {
         let filterStr = [];
-        searchFilters.forEach((value, key) => {
-            if (value.length !== 0) {
-                filterStr.push(encodeURIComponent(key) + "=" + encodeURIComponent(value));
+        searchFilters.forEach((values, key) => {
+            if (values.length > 0) {
+                for (let value of values) {
+                    filterStr.push(key + "=" + value);
+                }
             }
         });
-        return filterStr.join('&');
+        return filterStr.join(',');
     }
 }
 
--- a/ui/src/components/FacetsGrid.js
+++ b/ui/src/components/FacetsGrid.js
@@ -57,29 +57,32 @@ class FacetsGrid extends Component {
      * @param isSelected bool indicating whether this facetValue should be added to or removed from the query
      * */
     updateFacets(facetName, facetValue, isSelected) {
-        if (this.searchFilters.get(facetName) === undefined && isSelected) {
-            this.searchFilters.set(facetName, [facetValue]);
-        } else if (this.searchFilters.get(facetName) !== undefined) {
-            if (isSelected && !this.searchFilters.get(facetName).includes(facetValue)) {
-                this.searchFilters.get(facetName).push(facetValue);
-            } else if (!isSelected) {
-                this.searchFilters.set(facetName, this.removeFacet(this.searchFilters.get(facetName), facetValue));
+        let currFacetVals = this.searchFilters.get(facetName);
+        if (isSelected) {
+            // Add facetValue to the list of filters for facetName
+            if (currFacetVals === undefined) {
+                this.searchFilters.set(facetName, [facetValue]);
+            } else if (!this.searchFilters.get(facetName).includes(facetValue)) {
+                currFacetVals.push(facetValue);
             }
+        } else if (this.searchFilters.get(facetName) !== undefined) {
+            // Remove facetValue from the list of filters for facetName
+            this.searchFilters.set(facetName, this.removeFacet(currFacetVals, facetValue));
         }
-        this.api.facetsGet({filter: this.serialize(this.searchFilters)}, this.callback);
+        this.api.facetsGet({filter: this.serializeFilters(this.searchFilters)}, this.callback);
     }
 
     removeFacet(valueList, facetValue) {
-        for (let i = valueList.length-1; i >= 0; i--) {
-            if (valueList[i] === facetValue) {
-                valueList.splice(i, 1);
-                break;
+        let newValueList = [];
+        for (let i = 0; i < valueList.length; i++) {
+            if (valueList[i] !== facetValue) {
+                newValueList.push(valueList[i])
             }
         }
-        return valueList;
+        return newValueList;
     }
 
-    serialize(searchFilters) {
+    serializeFilters(searchFilters) {
         let filterStr = [];
         searchFilters.forEach((values, key) => {
             if (values.length > 0) {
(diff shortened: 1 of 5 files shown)


ui/src/components/FacetsGrid.js
```diff
@@ -1,26 +1,94 @@
...
+        this.searchFilters = new Map();
+        let apiClient = new ApiClient();
+        apiClient.basePath = '/api';
+        this.api = new FacetsApi(apiClient);
+        this.callback = function(error, data) {
+            if (error) {
+                console.error(error);
+                // TODO(alanhwang): Redirect to an error page
+            } else {
+                this.setState({facetsResponse: data});
```
melissachang: Looking at React examples, usually you extract things from response, eg from https://reactjs.org/docs/faq-ajax.html:
```
componentDidMount() {
    fetch("https://api.example.com/items")
      .then(res => res.json())
      .then(
        (result) => {
          this.setState({
            isLoaded: true,
            items: result.items
          });
        },
```
Conceptually, state is a list of facets. We don't care that we got the facets from an XHR.

ui/src/components/FacetsGrid.js
```diff
@@ -1,26 +1,94 @@
...
+        if (this.searchFilters.get(facetName) === undefined && isSelected) {
+            this.searchFilters.set(facetName, [facetValue]);
+        } else if (this.searchFilters.get(facetName) !== undefined) {
+            if (isSelected && !this.searchFilters.get(facetName).includes(facetValue)) {
+                this.searchFilters.get(facetName).push(facetValue);
+            } else if (!isSelected) {
+                this.searchFilters.set(facetName, this.removeFacet(this.searchFilters.get(facetName), facetValue));
+            }
+        }
+        console.log(this.serialize(this.searchFilters));
```
melissachang: Can this be deleted? This is easy to see from Network tab.

ui/src/components/FacetsGrid.js
```diff
@@ -1,26 +1,94 @@
...
+        for (let i = valueList.length-1; i >= 0; i--) {
+            if (valueList[i] === facetValue) {
+                valueList.splice(i, 1);
+                break;
+            }
+        }
+        return valueList;
+    }
+
+    serialize(searchFilters) {
```
melissachang: serializeFilters

ui/src/components/FacetsGrid.js
```diff
@@ -1,26 +1,94 @@
...
+    }
+
+    /**
+     * Updates the selection for a single facet value and refreshes the facets data from the server.
+     * @param facetName string containing the name of the facet corresponding to this value
+     * @param facetValue string containing the name of this facet value
+     * @param isSelected bool indicating whether this facetValue should be added to or removed from the query
+     * */
+    updateFacets(facetName, facetValue, isSelected) {
+        if (this.searchFilters.get(facetName) === undefined && isSelected) {
```
melissachang: I find this logic hard to follow. How about something like this?
```
if (isSelected) { // Checking check box
  if (this.searchFilters.get(facetName) === undefined)
    ...
  else
    ...
else // Unchecking check box
  ...
```

ui/src/components/FacetsGrid.js
```diff
@@ -1,26 +1,94 @@
...
+            } else if (!isSelected) {
+                this.searchFilters.set(facetName, this.removeFacet(this.searchFilters.get(facetName), facetValue));
+            }
+        }
+        console.log(this.serialize(this.searchFilters));
+        this.api.facetsGet({filter: this.serialize(this.searchFilters)}, this.callback);
+    }
+
+    removeFacet(valueList, facetValue) {
+        for (let i = valueList.length-1; i >= 0; i--) {
```
melissachang: This is easier for me to understand:
```
newValueList = []
for (let i = 0; i < valueList.length; i++) {
  if (valueList[i] !== facetValue) {
    newValueList.push(valueList[i])
  }
}
return newValueList;
```

ui/src/components/FacetsGrid.js
melissachang: This is what I see:

filter=Age=30-39,60-69&Region=southeast

This is what we want:

filter=Age=30-39&filter=Age=60-69&filter=Region=southeast

See https://swagger.io/docs/specification/serialization/#query

On second thought, probably this is better (up to you):

filter=Age=30-39,Age=60-69,Region=southeast

You would need to update api.yaml.

(Added this to things we should unit test in #7)

melissachang: Actually, we should be using Swagger-generated Facet, FacetValue, etc. Then we don't have to worry about serialization.

alahwa: Current implementation is your second suggestion.

Your idea to use the swagger-generated models is interesting, but I'm not sure how to do this practically. It would be nice if those came with handy serializing/deserializing utilities like protos, but currently when I encode something like a Gender facet with one value female I get `Gender&filter[values]=female`. As far as I can tell, it's not any kind of standard format that we can decode in Python, so I'm not sure how to go about deserializing this on the server side.

ui/src/components/FacetsGrid.js
melissachang: I mean, instead of
```
          this.setState({facetsResponse: data});
```
How about
```
          this.setState({
            totalCount: data.count,
            facets: data.facets
          });
```

ui/src/components/FacetsGrid.js
```diff
@@ -1,26 +1,98 @@
...
+    }
+
+    /**
+     * Updates the selection for a single facet value and refreshes the facets data from the server.
+     * @param facetName string containing the name of the facet corresponding to this value
+     * @param facetValue string containing the name of this facet value
+     * @param isSelected bool indicating whether this facetValue should be added to or removed from the query
+     * */
+    updateFacets(facetName, facetValue, isSelected) {
+        let currFacetVals = this.searchFilters.get(facetName);
```
melissachang: I'm ok with curr, but can this be currFacetValues? Technically Google Javascript [style guide says](https://google.github.io/styleguide/jsguide.html#naming-rules-common-to-all-identifiers) we shouldn't abbreviate.

ui/src/components/FacetsGrid.js
```diff
@@ -1,26 +1,98 @@
...
+        if (isSelected) {
+            // Add facetValue to the list of filters for facetName
+            if (currFacetVals === undefined) {
```
melissachang: I would expect either
```
            if (currFacetVals === undefined) {
                this.searchFilters.set(facetName, [facetValue]);
            } else {
                currFacetVals.push(facetValue);
            }
```
or
````
            if (currFacetVals === undefined) {
                this.searchFilters.set(facetName, [facetValue]);
            } else if (!this.searchFilters.get(facetName).includes(facetValue)) {
                currFacetVals.push(facetValue);
            } else {
                 show an error
            }
```

ui/src/components/FacetsGrid.js
melissachang: Oops, never mind about the swagger. I forgot Facet/FacetValue isn't being used for filtering.

ui/src/components/FacetsGrid.js
alahwa: Makes sense, done.

ui/src/components/FacetsGrid.js

ui/src/components/FacetsGrid.js
alahwa: Actually the former is fine because removeFacet will always iterate through the whole list anyway to remove potentially multiple copies of the same Facet in the worst case of us getting to this bad state.

PR #37, opened 2018-03-28
Show dataset name in UI

melissachang:
* Use convert_to_index_name() from data-explorer-indexers
* Add /api/dataset to API server which returns dataset name

datasetname


test/README.md
```diff
@@ -3,6 +3,6 @@ https://github.com/DataBiosphere/data-explorer-indexers
 
 This script loads test data into Elasticsearch, so one can quickly try out Data
 Explorer. Test data is from
-https://github.com/stedy/Machine-Learning-with-R-datasets/blob/master/insurance.csv
+[https://github.com/stedy/Machine-Learning-with-R-datasets/blob/master/insurance.csv](https://github.com/stedy/Machine-Learning-with-R-datasets/blob/master/insurance.csv)
```
alahwa: Maybe a more user-friendly string like `here`?

test/dataset.json
```diff
@@ -1,7 +1,9 @@
 {
-  // Elasticsearch index will be given this name.
-  "name": "test",
-  // Name of BigQuery field that is primary key. Elasticsearch documents will
-  // have this ID.
+  // Dataset name. This name will be displayed in the Data Explorer UI. Also,
+  // the Elasticsearch index will be given this name.
+  "name": "Insurance data",
+  // Name of BigQuery column that represents primary key. See
+  // https://github.com/DataBiosphere/data-explorer-indexers/tree/master/bigquery#overview
```
alahwa: I'm not sure how this link explains the primary key, and this will break if we ever decide to change that README without remembering to update this. An in-line comment description would probably be fine here.

ui/src/components/Header.js
```diff
@@ -1,24 +1,32 @@
 import './Header.css';
 
-import React from 'react';
-import AccountCircle from 'material-ui/svg-icons/action/account-circle.js';
+import React, { Component } from 'react';
 import {white} from 'material-ui/styles/colors';
 
-function Header(props) {
```
alahwa: I actually prefer the `function Header(props)` syntax to `class Header extends Component`. When I see the former defined as a function, I immediately understand that I'm looking at a stateless component, since `props` is always immutable. Since this class has no reason to hold `dataset` or `totalCount` as states, what do you think about returning to the original syntax and just calling using `const datasetName = props.datasetName` instead of holding these in states?

test/README.md
melissachang: Done

test/dataset.json
melissachang: It's from a [PR that's not merged yet](https://github.com/DataBiosphere/data-explorer-indexers/tree/config-info/bigquery#overview). It's a lengthy explanation.

This comment (lines 5 and 6) actually came from that PR. I figured I would update in this PR.

ui/src/components/Header.js
melissachang: Done

PR #38, opened 2018-03-28
Up the sleep time

alahwa:
Apparently the 2018 specialist eng laptops can't handle a 10sec latency.

Quick hack for now if it's not too disruptive for you, we should return to looking at a smarter implementation soon.

### DIFF
--- a/test/indexer.py
+++ b/test/indexer.py
@@ -29,7 +29,7 @@ def init_elasticsearch():
   # Wait for Elasticsearch to come up.
   # I had trouble getting wait_for_status to work, so just use sleep. See
   # https://discuss.elastic.co/t/cant-get-python-client-wait-for-status-to-work/123163
-  time.sleep(10)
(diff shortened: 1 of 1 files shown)


test/indexer.py
```diff
@@ -29,7 +29,7 @@ def init_elasticsearch():
   # Wait for Elasticsearch to come up.
   # I had trouble getting wait_for_status to work, so just use sleep. See
   # https://discuss.elastic.co/t/cant-get-python-client-wait-for-status-to-work/123163
-  time.sleep(10)
+  time.sleep(20)
```
melissachang: Can we do 15? Then we won't have to wait as long until we can use UI.

alahwa: Playing it dangerous but sure

alahwa:
Yes, everything works, although if I continue to run into these error's we'll have to up the sleep time again.

PR #40, opened 2018-03-28
Support for RangeFacet search

alahwa:
Extract the forward/backwards translation to their own functions to make it clear that these go together.

### DIFF
--- a/api/data_explorer/controllers/facets_controller.py
+++ b/api/data_explorer/controllers/facets_controller.py
@@ -30,8 +30,7 @@ def facets_get(filter=None):  # noqa: E501
                 #   name 10: count 15     (There are 15 people aged 10-19)
                 #   name 20: count 33     (There are 33 people aged 20-29)
                 # Convert "10" -> "10-19".
-                range_str = '%d-%d' % (name,
-                                       name + es_facet._params['interval'] - 1)
+                range_str = bucket_number_to_string(name, es_facet._params['interval'])
                 facet_values.append(FacetValue(name=range_str, count=count))
             else:
                 facet_values.append(FacetValue(name=name, count=count))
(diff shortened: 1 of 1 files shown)


api/data_explorer/controllers/facets_controller.py
```diff
@@ -51,9 +50,20 @@ def deserialize(filter_arr):
     for facet_filter in filter_arr:
         filter_str = urllib.unquote(facet_filter).decode('utf8')
         key_val = filter_str.split('=')
+        facet_name = key_val[0]
+        facet_value = key_val[1]
+        es_facet = current_app.config['ELASTICSEARCH_FACETS'][facet_name]
+        if isinstance(es_facet, HistogramFacet):
+            facet_value = bucket_string_to_number(facet_value)
         if len(key_val) == 2:
```
melissachang: Isn't len(key_val) always 2?

api/data_explorer/controllers/facets_controller.py
```diff
@@ -51,9 +50,20 @@ def deserialize(filter_arr):
     for facet_filter in filter_arr:
```
melissachang: Sorry, some minor comments I should have made on previous PR
```
# filter_str looks like "Gender=male"
for filter_str in filter_arr:
    filter_str = urllib.unquote(filter_str).decode('utf8')
```

api/data_explorer/controllers/facets_controller.py
```diff
@@ -51,9 +50,20 @@ def deserialize(filter_arr):
...
-            if not key_val[0] in parsed_filter:
-                parsed_filter[key_val[0]] = [key_val[1]]
+            if not facet_name in parsed_filter:
+                parsed_filter[facet_name] = [facet_value]
             else:
-                parsed_filter[key_val[0]].append(key_val[1])
+                parsed_filter[facet_name].append(facet_value)
     return parsed_filter
+
+def bucket_number_to_string(bucket_number, interval_size):
```
melissachang:
```
+def number_to_range(number, interval):
      """Converts "10" -> "10-20"."""
+    return '%d-%d' % (bucket_number, bucket_number + interval_size - 1)
+
+def range_to_number(range):
      """Converts "10-20" -> "10"."""
+    return int(bucket_string.split('-')[0]) 
```

api/data_explorer/controllers/facets_controller.py
alahwa: Done

api/data_explorer/controllers/facets_controller.py
alahwa: Done

api/data_explorer/controllers/facets_controller.py
alahwa: Yes, but since the 'name=value' syntax is just contractual between the server implementation and the client, there's nothing stopping anything from sending an incorrectly formatted filter string, so this is just a sanity check to make sure we only act if we're getting what we expected.

melissachang: Don't need to make any changes, just a comment in general -- it's a little weird to me to see "if len(key_val) == 2:" but not do anything if the condition isn't true. If you're going to make the check, throw an error if the check fails. (Or don't make the check.)

PR #42, opened 2018-03-28
Lift state from FacetsGrid to App

alahwa:
The end result is that App becomes the only stateful container and the FacetResponse counts correctly propagate to the Header. Note that as we continue to add logic, we might want to consider want a logic-containing component to reside under App that holds all of the FacetResponse-dependent components within it.

### DIFF
--- a/ui/src/App.js
+++ b/ui/src/App.js
@@ -13,12 +13,30 @@ class App extends Component {
       super(props);
       this.state = {
           datasetName: '',
-          facetsResponse: null
+          facets: null,
+          totalCount: null
       };
+
+      this.apiClient = new ApiClient();
+      this.apiClient.basePath = '/api';
+      this.facetsApi = new FacetsApi(this.apiClient);
+      this.facetsCallback = function(error, data) {
+          if (error) {
+              console.error(error);
+              // TODO(alanhwang): Redirect to an error page
+          } else {
+              this.setState({
+                  facets: data.facets,
+                  totalCount: data.count
+              });
+          }
+      }.bind(this);
+      this.searchFilters = new Map();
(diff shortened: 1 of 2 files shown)


ui/src/App.js
```diff
@@ -27,33 +45,23 @@ class App extends Component {
...
                     <Header
                         datasetName={this.state.datasetName}
-                        totalCount={this.state.facetsResponse.count}
+                        totalCount={this.state.totalCount}
                     />
-                    <FacetsGrid facetsResponse={this.state.facetsResponse} />
+                    <FacetsGrid
+                        updateFacets={this.updateFacets}
+                        facets={this.state.facets}
+                        totalCount={this.state.totalCount}/>
```
melissachang: From https://github.com/airbnb/javascript/tree/master/react#alignment
```
+                    <FacetsGrid
￼ +                        updateFacets={this.updateFacets}
￼ +                        facets={this.state.facets}
￼ +                        totalCount={this.state.totalCount}
                       />
```

ui/src/components/FacetsGrid.js
```diff
@@ -1,102 +1,32 @@
...
+    const facets = props.facets;
+    const totalCount = props.totalCount;
+    const updateFacets = props.updateFacets;
+    const facetsList = facets.map((facet) =>
+        <GridTile key={facet.name}>
+            <FacetCard
+                facet={facet}
+                count={totalCount}
+                updateFacets={updateFacets}
+                className='facetCard'/>
```
melissachang: `/>` on next line

ui/src/App.js
alahwa: Done

ui/src/components/FacetsGrid.js

PR #43, opened 2018-03-28
Fix a bug from swagger-codegen

alahwa:
The recent swagger-codegen update broke FacetedSearch. The API layer enforces that we provide queryParams as an array, not as comma separated values (which we were doing anyway) so we'll let the API layer handle that.

### DIFF
--- a/ui/src/App.js
+++ b/ui/src/App.js
@@ -31,6 +31,7 @@ class App extends Component {
               });
           }
       }.bind(this);
+      // Map of facetName:[facetValues] pairs
       this.searchFilters = new Map();
       this.updateFacets = this.updateFacets.bind(this);
   }
@@ -93,7 +94,7 @@ class App extends Component {
             // Remove facetValue from the list of filters for facetName
             this.searchFilters.set(facetName, this.removeFacet(currentFacetValues, facetValue));
         }
-        this.facetsApi.facetsGet({filter: this.serializeFilters(this.searchFilters)}, this.facetsCallback);
+        this.facetsApi.facetsGet({filter: this.filterMapToArray(this.searchFilters)}, this.facetsCallback);
     }
 
     removeFacet(valueList, facetValue) {
@@ -106,16 +107,22 @@ class App extends Component {
         return newValueList;
     }
 
-    serializeFilters(searchFilters) {
-        let filterList = [];
-        searchFilters.forEach((values, key) => {
+  /**
+   * Converts a Map of filters to an Array of filter strings interpretable by
+   * the backend
+   * @param filterMap Map of facetName:[facetValues] pairs
+   * @return [string] Array of "facetName=facetValue" strings
+   */
+    filterMapToArray(filterMap) {
+        let filterArray = [];
+        filterMap.forEach((values, key) => {
             if (values.length > 0) {
                 for (let value of values) {
-                    filterList.push(key + "=" + value);
+                    filterArray.push(key + "=" + value);
                 }
             }
         });
-        return filterList;
+        return filterArray;
     }
 }
 
--- a/ui/src/App.js
+++ b/ui/src/App.js
@@ -31,8 +31,8 @@ class App extends Component {
               });
           }
       }.bind(this);
-      // Map of facetName:[facetValues] pairs
-      this.searchFilters = new Map();
+      // Map from facet name to a list of facet values.
+      this.filterMap = new Map();
       this.updateFacets = this.updateFacets.bind(this);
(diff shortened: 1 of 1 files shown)


ui/src/App.js
```diff
@@ -107,15 +107,15 @@ class App extends Component {
     }
```
melissachang: Github won't let me leave a comment on line 34. "search" isn't used elsewhere, so can line 34 change to
```
// Map from facet name to list of facet values
this.filterMap = new Map();
```

ui/src/App.js
```diff
@@ -107,15 +107,15 @@ class App extends Component {
...
 
     serializeFilters(searchFilters) {
```
melissachang:
```
filterMapToArray()
   """Converts filter map to array of filter strings.

   Args:
     filterMap: map from filter name to list of filter values

   Returns:
     An array of filter strings, where each string looks like "Gender=male".
```

ui/src/App.js
```diff
@@ -107,15 +107,15 @@ class App extends Component {
...
-        let filterStr = [];
+        let filterList = [];
         searchFilters.forEach((values, key) => {
             if (values.length > 0) {
```
melissachang: Throw an error if values.length == 0, or don't check

ui/src/App.js
```diff
@@ -107,15 +107,15 @@ class App extends Component {
     }
 
     serializeFilters(searchFilters) {
-        let filterStr = [];
+        let filterList = [];
```
melissachang: filterArr or filterArray

Python has lists but I think these are arrays in js

ui/src/App.js
```diff
@@ -107,15 +107,15 @@ class App extends Component {
...
                 for (let value of values) {
-                    filterStr.push(key + "=" + value);
+                    filterList.push(key + "=" + value);
                 }
             }
         });
-        return filterStr.join(',');
+        return filterList;
     }
 }
```
melissachang: Adding comment here so you can reply to it if you want.

I git checkout and I see this bug:
* Check a facet. Things work as expected.
* Uncheck same facet. Get console error and UI doesn't update.

ui/src/App.js
alahwa: This is not an error case, check > uncheck will result in an initialized key corresponding to an empty array, in which case we'll just ignore it rather than provide an empty filter.

ui/src/App.js
alahwa: This is a separate issue which I have a followup PR staged for.

ui/src/App.js
alahwa: Done

ui/src/App.js
alahwa: Done

ui/src/App.js
alahwa: Done

ui/src/App.js
```diff
@@ -31,6 +31,7 @@ class App extends Component {
               });
           }
       }.bind(this);
+      // Map of facetName:[facetValues] pairs
```
melissachang: "pairs" is confusing -- where is the pair?

// Map from facet name to a list of facet values. 

Same on line 113

ui/src/App.js
```diff
@@ -31,6 +31,7 @@ class App extends Component {
...
       this.searchFilters = new Map();
```
melissachang: Can searchFilters be renamed to filterMap?

ui/src/App.js
```diff
@@ -106,16 +107,22 @@ class App extends Component {
...
+  /**
+   * Converts a Map of filters to an Array of filter strings interpretable by
+   * the backend
+   * @param filterMap Map of facetName:[facetValues] pairs
+   * @return [string] Array of "facetName=facetValue" strings
+   */
+    filterMapToArray(filterMap) {
+        let filterArray = [];
+        filterMap.forEach((values, key) => {
             if (values.length > 0) {
```
melissachang: // Scenario where there are no values for a key:
// A single value is checked for a facet. The value is unchecked.
// The facet name will still be a key in filterMap, but there will be no values.
if (values.length > 0) {

ui/src/App.js
alahwa: Done

ui/src/App.js
alahwa: Done

ui/src/App.js
alahwa: Done

PR #46, opened 2018-03-29
Convert the header to a toolbar

alahwa:
Center the elements in the header by switching to use a material toolbar instead.

screen shot 2018-03-30 at 5 19 21 pm

### DIFF
--- a/ui/src/components/Header.css
+++ b/ui/src/components/Header.css
@@ -1,41 +1,28 @@
-.header {
-    width: 45%;
-    margin-top: 15px;
-    vertical-align: center;
-}
-
-.headerBox {
-    width: 100%;
+.toolbar {
     background-color: #553A70 !important;
-    margin-bottom: 20px;
-    overflow: hidden;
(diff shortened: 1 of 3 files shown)


ui/src/components/Header.css
```diff
@@ -1,41 +1,28 @@
-.header {
-    width: 45%;
-    margin-top: 15px;
-    vertical-align: center;
-}
-
-.headerBox {
-    width: 100%;
+.toolbar {
```
melissachang: Can this be a [FAB (floating action button)](https://material.io/guidelines/components/buttons-floating-action-button.html)? You don't have to get the floating stuff working, obviously, but I would like a red button on lower right, [similar to Inbox](https://lh3.googleusercontent.com/bCf1aJaKy9CG4W8viDPPEKfwW0aE_twWRR7acRL21fJFlBsNLKT2rAKO1uxxRC5daQ=h310).

alahwa: Obsolete

PR #47, opened 2018-03-29
Gray out non-selected FacetValues

alahwa:
screen shot 2018-03-29 at 4 34 41 pm

### DIFF
--- a/ui/src/components/FacetCard.css
+++ b/ui/src/components/FacetCard.css
@@ -19,7 +19,7 @@
 
 .subHeader {
     margin-top: 5px;
-    color: #553A70;
+    color: gray;
 }
 
 .totalCount {
@@ -31,5 +31,5 @@
 }
 
 .grayText {
-    color: gray !important;
+    color: silver !important;
 }
--- a/ui/src/components/FacetCard.css
+++ b/ui/src/components/FacetCard.css
@@ -19,7 +19,7 @@
 
 .subHeader {
     margin-top: 5px;
-    color: gray;
+    color: #553A70;
 }
 
 .totalCount {
@@ -28,4 +28,8 @@
 
 .numberSelected {
     float: right;
+}
+
+.grayText {
+    color: gray !important;
 }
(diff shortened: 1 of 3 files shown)


ui/src/components/FacetCard.css
```diff
@@ -29,3 +29,7 @@
 .numberSelected {
     float: right;
 }
+
+.grayText {
+    color: gray !important;
```
melissachang: I think `silver` looks better.

ui/src/components/FacetCard.js
```diff
@@ -1,37 +1,68 @@
...
+        if (isInputChecked) {
+            newValues.push(facetValue.name);
+        } else {
+            newValues.splice(newValues.indexOf(facetValue.name), 1);
+        }
+        this.setState({selectedValues: newValues});
+        this.props.updateFacets(this.props.facet.name, facetValue.name, isInputChecked);
+    }
+
+    isUnselected(facetValue) {
```
melissachang: isSelected() seems more natural to me

ui/src/components/FacetCard.css
```diff
@@ -19,7 +19,7 @@
 
 .subHeader {
     margin-top: 5px;
-    color: gray;
+    color: #553A70;
```
melissachang: I prefer gray over purple. If it's purple, it looks like the most important part of the card.

alahwa: Done

ui/src/components/FacetCard.css
alahwa: Done

ui/src/components/FacetCard.js
alahwa: The set of not unselected things actually consists of  both selected values and all values in facets which have no selected values. I chose this naming using the negation of those things in order to avoid that messiness. And functionally, we don't have any need to negate `isUnselected`, so the syntax would be `~isSelected` anyway.

PR #64, opened 2018-04-20
Fix flakiness with end-to-end test

melissachang:
Occasionally test failed with:
```
 FAIL  tests/e2e/test.js
  End-to-end
    ✓ Header (484ms)
    ✓ Age facet (6ms)
    ✕ Faceted search (48ms)

  ● End-to-end › Faceted search

    expect(received).toBe(expected) // Object.is equality
    
    Expected value to be:
      "137"
    Received:
      "1338"

      31 |   async function assertHeaderTotalCount(count) {
      32 |     const totalCount = await page.$eval("div.totalCountText", e => e.
innerText);
    > 33 |     await expect(totalCount).toBe(count);
      34 |   }
      35 | 
      36 |   async function assertFacet(
      
      at _callee5$ (tests/e2e/test.js:33:36)
      at tryCatch (node_modules/regenerator-runtime/runtime.js:62:40)
      at Generator.invoke [as _invoke] (node_modules/regenerator-runtime/runtime
.js:296:22)
      at Generator.prototype.(anonymous function) [as next] (node_modules/regene
rator-runtime/runtime.js:114:21)
      at step (tests/e2e/test.js:1:225)
      at tests/e2e/test.js:1:385
```
I was able to repro consistently by putting `import time; time.sleep(1)` into `facets_controller.py facets_get()`.

### DIFF
--- a/ui/tests/e2e/test.js
+++ b/ui/tests/e2e/test.js
@@ -1,7 +1,7 @@
 describe("End-to-end", () => {
   beforeAll(async () => {
-  await page.goto("http://localhost:4400");
-  await page.waitForSelector("span.datasetName")
+    await page.goto("http://localhost:4400");
+    await page.waitForSelector("span.datasetName");
   });
 
   test("Header", async () => {
@@ -19,7 +19,7 @@ describe("End-to-end", () => {
     await facetValueRow.click("input");
     // After data is returned from backend, Smoker facet will have "107".
     // See #63 for why we can't simply wait for div.grayText.
-    await page.waitForXPath("//text()[contains(.,'107')]")
+    await page.waitForXPath("//text()[contains(.,'107')]");
 
     // Assert page updated correctly
     await assertHeaderTotalCount("137");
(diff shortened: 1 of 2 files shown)


ui/tests/e2e/test.js
```diff
@@ -16,6 +17,9 @@ describe("End-to-end", () => {
     // Click first Age facet value
     let facetValueRow = await getFacetValueRow("Age", "10-19");
     await facetValueRow.click("input");
+    // After data is returned from backend, Smoker facet will have "107".
```
alahwa: For `expect(grayDiv).toBeTruthy();`, is this assertion necessary? Won't the test fail anyway if that div doesn't become gray?

ui/tests/e2e/test.js
```diff
@@ -16,6 +17,9 @@ describe("End-to-end", () => {
...
+    // See #63 for why we can't simply wait for div.grayText.
+    await page.waitForXPath("//text()[contains(.,'107')]");
```
alahwa: Instead of looking for a random value that's present in the new xpath but not the old one, we could `await totalHeaderCount = 137` and have that serve as the assertion.

ui/tests/e2e/test.js
melissachang: How would the test fail? If there's a bug where div doesn't become gray, grayDiv will be null and test will pass.

ui/tests/e2e/test.js
melissachang: Done

PR #71, opened 2018-05-11
Initial yapf lint + CircleCI config

alahwa:
Set up a CircleCI config to verify that builds all pass python linting. Also make a first pass running yapf to get the code base to a properly formatted state (next PR will verify that the starting build is linted properly and then runs the formatter only over staged files).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,7 +28,7 @@ jobs:
       - run:
           name: Run the yapf python linter
           command: |
-            pip install tox yapf
+            pip install yapf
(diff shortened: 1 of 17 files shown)


.circleci/config.yml
```diff
@@ -26,7 +26,12 @@ jobs:
           paths:
             - "ui/node_modules"
       - run:
-          name: Run the linter
+          name: Run the yapf python linter
+          command: |
+            pip install tox yapf
```
melissachang: tox is needed?

PR #72, opened 2018-05-11
Add a yapf precommit hook

alahwa:


hooks/pre-commit
```diff
@@ -13,3 +13,25 @@ then
   echo Please commit to a working branch and push it to GitHub for review/merge.
   exit 1
 fi
+
+readonly changed_files_command="git diff --cached --name-only | grep '\.py$' | grep -v 'servers/[^\/]*/jobs/models' | grep -v 'setup.py'"
+readonly files_in_commit=$(eval $changed_files_command)
+echo $files_in_commit
```
melissachang: I prefer not to have all these echo lines (line 19, 22, 28) for every commit.

Ideally precommit would automatically format and save files, without manual intervention, like "pretty-quick --staged". Since this PR doesn't do that, I would like to either cancel this PR, or remove all this logging.

alahwa: Done

hooks/pre-commit
```diff
@@ -13,3 +13,22 @@ then
   echo Please commit to a working branch and push it to GitHub for review/merge.
   exit 1
 fi
+
+readonly changed_files_command="git diff --cached --name-only | grep '\.py$' | grep -v 'servers/[^\/]*/jobs/models' | grep -v 'setup.py'"
```
melissachang: Why do you exclude these files, like setup.py? The list of files to exclude should be kept in sync with CircleCI. CircleCI only excludes ui/node_modules.

hooks/pre-commit
```diff
@@ -13,3 +13,22 @@ then
...
+if [ -n "$files_in_commit" ]
+then
+  # Stash un-staged changes to ensure that the staged changes pass the linter
+  git stash -q --keep-index
+  readonly yapf_out=$(yapf -d $files_in_commit)
+
+  if ! [ -z "$yapf_out" ]
+  then
+    echo "Reformatting staged files, please review them and stage the changes."
+    yapf -i $files_in_commit --exclude ui/node_modules/
```
melissachang:
```
# Keep in sync with .circleci/config.yml
```
Also add comment to .circleci/config.yml

hooks/pre-commit
```diff
@@ -13,3 +13,22 @@ then
...
+readonly files_in_commit=$(eval $changed_files_command)
+if [ -n "$files_in_commit" ]
+then
+  # Stash un-staged changes to ensure that the staged changes pass the linter
+  git stash -q --keep-index
+  readonly yapf_out=$(yapf -d $files_in_commit)
+
+  if ! [ -z "$yapf_out" ]
+  then
+    echo "Reformatting staged files, please review them and stage the changes."
```
melissachang: I just tested this out.
- Made a change to a file that would cause yapf to fail
- git commit -m 'blah'
- commit went through with incorrectly formatted file
- After commit, file is still incorrectly formatted.

That's not what you expect, right?

My advice is to abandon this PR, rather than try to get this working.

If you can get the automatic reformatting working, then this PR is worth it. Seems doable given you have file_in_commit.

melissachang:
Facebook is working on a new Python formatter, [Black](https://github.com/ambv/black), which they will use for their .py files. I think Black will have more adoption than YAPF over time. I prefer Black over YAPF because:
- [Black is much faster](https://news.ycombinator.com/item?id=17155285)
- YAPF is non-determinstic; Black doesn't have this problem. (YAPF-ing a file twice can produce different results from YAPF-ing once.)
- [Black README describes easy precommit integration](https://github.com/ambv/black#editor-integration), which YAPF doesn't support. (I assume the Black integration supports automatically restaging formatted files; this would have to be verified.)
- [Black README describes easy .vim integration.](https://github.com/ambv/black#vim), which YAPF doesn't suport (I run into [this bug](https://github.com/mindriot101/vim-yapf/issues/5)).

What do you think about looking into Black?

PR #79, opened 2018-06-07
 Include bq table references in the /dataset api def

alahwa:
This array will be parsed from the config and provided to the ui for the export to firecloud functionality


api/api.yaml
```diff
@@ -49,6 +49,10 @@ definitions:
     properties:
       name:
         type: string
+      tableReferences:
```
melissachang: Let's do `tableNames`

PR #80, opened 2018-06-08
Instructions for running swagger-codegen for macOS

alahwa:


melissachang:
Did you see the email that CircleCI is broken on master? Can you fix that in this PR?

alahwa:
Should be working now

PR #81, opened 2018-06-08
For dataset_config dir, use symlink instead of copy files

melissachang:
With copied files, it's confusing that there are two sets of files.

Also add a clarification, requested by Randall, that Dataset owner must be in the loop when picking a Google Group for IAP.


deploy/README.md
```diff
@@ -25,22 +21,33 @@ to bring up Elasticsearch in GKE and index dataset.
...
+  yet.
   * In an incognito window, navigate to `https://PROJECT_ID.appspot.com`. Login
   as a user who is not in the Google Group. You should see a "You don't have
   access" page.
 
 ### Deploy the API Server
 
-* Ensure that `api/dataset_config/` contains your config. If you are using the
-default platinum_genomes dataset, you can get config files [from here](https://github.com/DataBiosphere/data-explorer-indexers/tree/master/bigquery/config/platinum_genomes).
+* Make sure that `api/dataset_config/current` is a symlink to your dataset
```
alahwa: `is a symlink` > `contains`. We recommend symlinking in our setup instructions, however this isn't a requirement and users are free to provide either a symlink or hard files here, as long as current/ contains the config.

melissachang: Done

PR #84, opened 2018-06-12
Use DATASET_CONFIG_DIR instead of current symlink

melissachang:
current symlink works with local deployment, but doesn't work with App Engine deployment. In general, [App Engine doesn't support symlinks in docker build context](https://groups.google.com/d/msg/google-cloud-sdk/2fmY1uHNziY/9CGvPsfJCAAJ).

I think this is cleaner anyways.

Confirmed that these work:
* `docker-compose up --build`
* `DATASET_CONFIG_DIR=/app/dataset_config/platinum_genomes docker-compose up --build`
* App Engine deployment

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,28 +5,29 @@
 
 ## Quickstart
 
-Run Data explorer with a test dataset:
+Run local Data explorer with a test dataset:
 
 * `docker-compose up --build`
 * Navigate to `localhost:4400`
 
-To use a different dataset:
+## Run local Data explorer with a specific dataset
 
-* For each dataset, there is a directory in `api/dataset_config`, e.g.
-  `api/dataset_config/amp_pd`.
-    * If you used https://github.com/DataBiosphere/data-explorer-indexers to
-      index your dataset, we recommend moving your dataset config directory from
-      that repo to `api/dataset_config/MY_DATASET`. If you need to run
-      more commands in the `data-explorer-indexers` repo, you can simply refer
-      to `api/dataset_config/MY_DATASET` in this repo.
-* Index your data into Elasticsearch. You can use one of the indexers at
-  https://github.com/DataBiosphere/data-explorer-indexers, or any other indexer.
-  https://github.com/DataBiosphere/data-explorer-indexers explains how to
-  set up the config files. Specifically:
-  * There must be a file named `dataset.json` that has a `name` field. This
-    determines the name of the Elasticsearch index.
-  * There must be a file named `facet_fields.csv` with the `readable_field_name`
-    column filled out.
+This repo contains an API server that reads from Elasticsearch, and a UI server
+that calls the API server. When running the UI and API servers, it is assumed
+that the dataset has already been indexed into Elasticsearch.
+
+To index your dataset into Elasticsearch, you can use one of the indexers at
+https://github.com/DataBiosphere/data-explorer-indexers, or any other indexer.
+
+* Index your data into Elasticsearch.
+* Create a directory in `api/dataset_config`, e.g. `api/dataset_config/amp_pd`.
+    * If you used https://github.com/DataBiosphere/data-explorer-indexers, copy
+    the config directory from there.
+    * If you used your own indexer, the config files must follow
+    this format ([see examples](https://github.com/DataBiosphere/data-explorer-indexers/tree/master/bigquery/dataset_config/platinum_genomes)):
+      * `dataset.json` has a `name` field. This determines the name of the Elasticsearch index.
+      * `facet_fields.csv` has the `readable_field_name` column filled out.
+
 * `DATASET_CONFIG_DIR=/app/dataset_config/<my dataset> docker-compose up --build`
 * Navigate to `localhost:4400`
 
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ To index your dataset into Elasticsearch, you can use one of the indexers at
 https://github.com/DataBiosphere/data-explorer-indexers, or any other indexer.
 
 * Index your data into Elasticsearch.
-* Create a directory in `api/dataset_config`, e.g. `api/dataset_config/amp_pd`.
+* Create `api/dataset_config/<my dataset>`.
     * If you used https://github.com/DataBiosphere/data-explorer-indexers, copy
     the config directory from there.
     * If you used your own indexer, the config files must follow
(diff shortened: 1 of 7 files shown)


api/app.yaml
```diff
@@ -6,4 +6,4 @@ entrypoint: gunicorn -b ":$PORT" data_explorer.__main__:app
 env_variables:
   PATH_PREFIX: /api
   ELASTICSEARCH_URL: http://ELASTICSEARCH_URL:9200/
-  DATASET_CONFIG_DIR: /app/dataset_config
+  DATASET_CONFIG_DIR: /app/dataset_config/platinum_genomes
```
alahwa: This deployment config should be unset rather than pointing to test data by default.

deploy/README.md
```diff
@@ -40,14 +40,13 @@ domain name: `https://PROJECT_ID.appspot.com`
 
 ### Deploy the API Server
 
-* Make sure that `api/dataset_config/current` contains your dataset config
```
alahwa: This bullet point and subpoints could be reduced down to:

* Set DATASET_CONFIG_DIR in `app.yaml`

With an optional subpoint for copying platinum genomes config for test data.

api/app.yaml
melissachang: Why -- everywhere else in this and data-explorer-indexer repo, we use a test dataset by default.

Also, I like having a sample string here. Otherwise, it's not clear that it should start with "/app/dataset_config". And it's not clear if this should be "/app/dataset_config" or "/app/dataset_config/platinum_genomes".

deploy/README.md
melissachang: Done

melissachang:
I changed the README as we discussed. Hopefully the "Run local Data explorer with a specific dataset" section is more clear. PTAL.

README.md
```diff
@@ -5,32 +5,30 @@
...
+
+This repo contains an API server that reads from Elasticsearch, and a UI server
+that calls the API server. When running the UI and API servers, it is assumed
+that the dataset has already been indexed into Elasticsearch.
+
+To index your dataset into Elasticsearch, you can use one of the indexers at
+https://github.com/DataBiosphere/data-explorer-indexers, or any other indexer.
+
+* Index your data into Elasticsearch.
+* Create a directory in `api/dataset_config`, e.g. `api/dataset_config/amp_pd`.
```
dionloy: * Create `api/dataset_config/<my dataset>`

(to match the similar style below).

README.md
```diff
@@ -5,28 +5,29 @@
...
-      to `api/dataset_config/MY_DATASET` in this repo.
-* Index your data into Elasticsearch. You can use one of the indexers at
-  https://github.com/DataBiosphere/data-explorer-indexers, or any other indexer.
-  https://github.com/DataBiosphere/data-explorer-indexers explains how to
-  set up the config files. Specifically:
-  * There must be a file named `dataset.json` that has a `name` field. This
-    determines the name of the Elasticsearch index.
-  * There must be a file named `facet_fields.csv` with the `readable_field_name`
-    column filled out.
+## Run local Data explorer with a specific dataset
```
alahwa: Can we have two separate sections under this? One for setting up data explorer using data-explorer-indexers, and one for setting up data explorer on top of any existing elasticsearch server?

melissachang: 3 out 4 steps are identical, regardless of which indexer you use. Rather than duplicate those 3 steps, I think this way is cleaner. I don't think it's a burden to the user that step 2 has 2 bullets. Let's discuss next time you're in the office. :)

README.md
```diff
@@ -5,32 +5,30 @@
...
+that the dataset has already been indexed into Elasticsearch.
+
+To index your dataset into Elasticsearch, you can use one of the indexers at
+https://github.com/DataBiosphere/data-explorer-indexers, or any other indexer.
+
+* Index your dataset into Elasticsearch.
+* Create `api/dataset_config/<my dataset>`
+    * If you used https://github.com/DataBiosphere/data-explorer-indexers, copy
+    the config directory from there.
+    * If you used your own indexer, the config files must follow
```
alahwa: I'm not exactly sure what this point adds to the self-documenting config templates. If the purpose is to describe the minimum set of fields that need to be filled out, we could say:

* Otherwise, your config files must, at minimum, contain non-null values for:
  * The `name` field in `dataset.json`
  * The `readable_field_name` column in `facet_fields.csv`

README.md
```diff
@@ -5,32 +5,30 @@
...
-  https://github.com/DataBiosphere/data-explorer-indexers explains how to
-  set up the config files. Specifically:
-  * There must be a file named `dataset.json` that has a `name` field. This
-    determines the name of the Elasticsearch index.
-  * There must be a file named `facet_fields.csv` with the `readable_field_name`
-    column filled out.
-* `docker-compose up --build`
+## Run local Data explorer with a specific dataset
+
+This repo contains an API server that reads from Elasticsearch, and a UI server
```
alahwa: This bit is describing how our code works, so would probably belong better in the Architecture section so we can keep this to just a set of instructions. Unless you want to make a comment about how our docker-compose can use an existing Elasticsearch instance as long as you provide the config, and defaults to a test dataset otherwise. That would be fine here.

alahwa: Alternatively, if you wanted this to be a sort of one liner overview, it might be good to put something similar to this at the top of the README:

The Data Explorer is an interface for easily visualizing and examining data from an Elasticsearch index.

README.md
alahwa: That makes sense, we should keep it as is then. I made some additional suggestions aimed at clarifying some of the parts I found confusing

README.md
```diff
@@ -5,32 +5,30 @@
```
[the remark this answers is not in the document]
melissachang: I'd rather leave it towards the bottom, because it won't be useful most of the time. I'm not sure changing vm.max_map_count is needed for every machine. Someone doesn't need to install Swagger if they just want to run local Data Explorer.

README.md
melissachang: Done, with some edits to your suggested wording.

README.md
melissachang: Rephrased.

README.md
```diff
@@ -5,32 +5,27 @@
...
-  https://github.com/DataBiosphere/data-explorer-indexers explains how to
-  set up the config files. Specifically:
-  * There must be a file named `dataset.json` that has a `name` field. This
-    determines the name of the Elasticsearch index.
-  * There must be a file named `facet_fields.csv` with the `readable_field_name`
-    column filled out.
-* `docker-compose up --build`
+## Run local Data Explorer with a custom dataset
+
+Before you can run the servers in this repo to display a Data Explorer UI,
```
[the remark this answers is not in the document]
melissachang: I'm trying to give context on why one needs to index data into Elasticsearch. I restructured, PTAL.

PR #88, opened 2018-06-20
Turn off service worker caching since it doesn't work with IAP

melissachang:
Confirmed Baseline DE now works. Also redeploying test, and AMP PD DE.

### DIFF
--- a/ui/src/index.js
+++ b/ui/src/index.js
@@ -5,4 +5,21 @@ import App from "./App";
 import registerServiceWorker from "./registerServiceWorker";
 
 ReactDOM.render(<App />, document.getElementById("root"));
-registerServiceWorker();
+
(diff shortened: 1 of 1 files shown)


ui/src/index.js
```diff
@@ -5,4 +5,21 @@ import App from "./App";
 import registerServiceWorker from "./registerServiceWorker";
 
 ReactDOM.render(<App />, document.getElementById("root"));
-registerServiceWorker();
+
```
alahwa: This seems a bit excessive for code comments, most of this context could probably just go in the PR description.

melissachang: Done

melissachang:
FYI - Now that this [IAP issue has been fixed](https://issuetracker.google.com/issues/119581489), this PR might not be needed anymore.

PR #89, opened 2018-06-20
Automate API server deployment

melissachang:
Tested with test-data-explorer.appspot.com

After this PR, I'll add to data-explorer-indexers repo: bigquery/dataset_config/template/deploy.json

I also improved the error message when Elasticsearch index is missing. Namely, the error message prints index name.

### DIFF
--- a/api/data_explorer/dataset_faceted_search.py
+++ b/api/data_explorer/dataset_faceted_search.py
@@ -87,7 +87,8 @@ def get_facets():
 
     using = Elasticsearch(current_app.config['ELASTICSEARCH_URL'])
     try:
-        current_app.config['INDEX_NAME'], 'type', using=using).to_dict()
+        mapping = Mapping.from_es(
+            current_app.config['INDEX_NAME'], 'type', using=using).to_dict()
     except TransportError as e:
         if 'index_not_found_exception' in e.error:
             current_app.logger.error('Index %s not found at %s' %
--- a/api/data_explorer/dataset_faceted_search.py
+++ b/api/data_explorer/dataset_faceted_search.py
@@ -94,7 +94,7 @@ def get_facets():
             current_app.logger.error('Index %s not found at %s' %
                                      (current_app.config['INDEX_NAME'],
                                       current_app.config['ELASTICSEARCH_URL']))
-            return []
+            raise e
 
     # Preserve order, so facets are returned in the same order as facet_fields.csv
     facets = OrderedDict()
--- a/api/data_explorer/dataset_faceted_search.py
+++ b/api/data_explorer/dataset_faceted_search.py
@@ -9,6 +9,7 @@ from flask import current_app
 
 from collections import OrderedDict
 from elasticsearch import Elasticsearch
+from elasticsearch.exceptions import TransportError
 from elasticsearch_dsl import HistogramFacet
 from elasticsearch_dsl import FacetedSearch
 from elasticsearch_dsl import Mapping
@@ -85,8 +86,15 @@ def get_facets():
     """
 
     using = Elasticsearch(current_app.config['ELASTICSEARCH_URL'])
-    mapping = Mapping.from_es(
+    try:
         current_app.config['INDEX_NAME'], 'type', using=using).to_dict()
+    except TransportError as e:
+        if 'index_not_found_exception' in e.error:
+            current_app.logger.error('Index %s not found at %s' %
+                                     (current_app.config['INDEX_NAME'],
+                                      current_app.config['ELASTICSEARCH_URL']))
+            return []
+
     # Preserve order, so facets are returned in the same order as facet_fields.csv
     facets = OrderedDict()
     for facet_row in get_facet_rows():
--- a/deploy/deploy-api.sh
+++ b/deploy/deploy-api.sh
@@ -1,6 +1,12 @@
 #!/bin/bash
 #
 # Deploy API Server on GCP.
+#
+# This script assumes Elasticsearch was deployed following the instructions in
+# https://github.com/DataBiosphere/data-explorer-indexers/. Specifically, there
+# must be a GKE cluster named "elasticsearch-cluster".
+#
+# jq, gcloud, and kubectl must be installed before running this script.
 
 if (( $# != 1 ))
 then
@@ -11,7 +17,7 @@ then
 fi
 
 dataset=$1
-project_id=$(grep -v "//" api/dataset_config/${dataset}/deploy.json | jq --raw-output '.project_id')
+project_id=$(jq --raw-output '.project_id' api/dataset_config/${dataset}/deploy.json)
 
 echo "Deploying ${dataset} API Server to project ${project_id}"
 echo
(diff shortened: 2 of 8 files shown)


deploy/deploy-api.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
```
alahwa: Document the list of assumptions both for the user and for us if we make changes to our deployment stack that would potentially break this: cluster running on GKE named elasticsearch-cluster, kubectl and gcloud installed and configured globally, anything else?

deploy/deploy-api.sh
```diff
@@ -0,0 +1,33 @@
...
+if (( $# != 1 ))
+then
+  echo "Usage: deploy/deploy-api.sh <dataset>"
+  echo "  where <dataset> is the name of a directory in api/dataset_config/"
+  echo "Run this script from project root"
+  exit 1
+fi
+
+dataset=$1
+project_id=$(grep -v "//" api/dataset_config/${dataset}/deploy.json | jq --raw-output '.project_id')
```
alahwa: This line prints: `grep: api/dataset_config/platinum_genomes/deploy.json: No such file or directory`

I'm not sure what `deploy.json` is, what's this line doing?

deploy/deploy-api.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+#
+# Deploy API Server on GCP.
+
+if (( $# != 1 ))
+then
+  echo "Usage: deploy/deploy-api.sh <dataset>"
```
alahwa: Add a `./`

deploy/deploy-api.sh
alahwa: "^//"?

melissachang: deploy.json looks like:
```
{
  // GCP project to deploy to
  "project_id": "test-data-explorer"
}
```
After this PR is in, I will add to it dataset configs in data-explorer-indexer, so in the future, when user runs this script, it will exist

melissachang: I get:
```
parse error: Invalid numeric literal at line 2, column 5
```

deploy/deploy-api.sh
melissachang: Not needed. In the past when I requested this, it was needed

api/data_explorer/dataset_faceted_search.py
```diff
@@ -85,8 +86,16 @@ def get_facets():
...
-        current_app.config['INDEX_NAME'], 'type', using=using).to_dict()
+    try:
+        mapping = Mapping.from_es(
+            current_app.config['INDEX_NAME'], 'type', using=using).to_dict()
+    except TransportError as e:
+        if 'index_not_found_exception' in e.error:
+            current_app.logger.error('Index %s not found at %s' %
+                                     (current_app.config['INDEX_NAME'],
+                                      current_app.config['ELASTICSEARCH_URL']))
+            return []
```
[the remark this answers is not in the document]
alahwa: `return []` > `raise e` should follow the normal program logic with the addition of the logging

deploy/deploy-api.sh
alahwa: Example: I decide to rename `elasticsearch-cluster` to `es-cluster`, I update the data-explorer-indexers repo accordingly. Result is that this script breaks

Since the nature of this is that it will be fairly fragile against changes like that, documenting exactly what the assumptions are will be useful to prevent introducing bugs.

deploy/deploy-api.sh
alahwa: That's fine, in general it would be nice to order these PRs such that they aren't reliant on future changes.

alahwa: Maybe escape the forward slashes? The gist of the comment was to anchor the regex to only `grep -v` lines that start with `//`

api/data_explorer/dataset_faceted_search.py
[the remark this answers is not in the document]
melissachang: Done

deploy/deploy-api.sh
melissachang: Well, documenting here won't be useful to someone changing data-explorer-indexers repo.

Eventually we'll have CD. If someone makes a breaking change like that, we will get a CD failure email.

deploy/deploy-api.sh
melissachang: FYI - I wanted to do this PR first, since this PR establishes the automation, in case that changed something about deploy.json. (Note in the PR description, I said I would add deploy.json to data-explorer-indexers repro after this PR is in.)

melissachang: Actually, I realized this doesn't make sense. Say the json file has:
```
  "project_id": "test-data-explorer"   // some comment
```
You're suggesting I leave that line in. But then jq will fail because the comment isn't valid json.

The right approach would be to strip only the comment. I don't think it's worth the effort to figure that out. (Note: jsmin handles this case correctly, but I don't want to make the user install jsmin.)

alahwa: This overhead is why I initially pushed to not have comments in our json. However, if we allow comments here, we should handle this case, because if we choose to leave the line in or remove it totally, our program will fail (in a very opaque way, as a user I'd have a very tough time tracking down `grep -v "//"` as the root cause of my `missing project_id` error). However, I also agree it's not worth it to write our own json parser. What do you think about changing our json config to yaml instead? That supports comments and is generally easier on the eyes when it comes to human readability.

deploy/deploy-api.sh
alahwa: If you're heavily against the documentation here I won't block you on it. I still think it would be useful to list assumptions to this script for the same reasons why we always document public methods with what kinds of inputs are acceptable.

deploy/deploy-api.sh
melissachang: I prefer json because it's widely used within the community. How about I add a comment to deploy.json saying that comments must begin at the beginning of a line?

deploy/deploy-api.sh
melissachang: Done

deploy/deploy-api.sh
alahwa: I think regardless of where we settle we should be consistent throughout the way we treat comments in json. So if we remove only whole line comments here, we should do the same in our indexers, etc.

Is it really worth having comments in our json to the point where we need to write comments about our comment structure? This seems like it's adding a fair amount of complexity on this:
```
{
  "project_id": "GCP_PROJECT_ID"
}
```
To me, this seems sufficiently self explanatory without the need for comments. I'm not sure exactly what you had in mind for deploy.json, but if it's just the project_id there's other ways to go about it and I'm not sure we need a separate config file for that.d

melissachang: I'll remove the comment from deploy.json in next PR. If I want to add a comment later, let's discuss in person.

alahwa: `-v "//"` no longer required here, at least for the time being

melissachang: Done. Oops, forgot to push some previous fixes, pushed now

PR #93, opened 2018-07-03
Use ui.json instead of facet_fields.csv

melissachang:
Corresponds to https://github.com/DataBiosphere/data-explorer-indexers/pull/35

Tested:
- Local default dataset
- Local Platinum Genomes


api/dataset_config/platinum_genomes/README.md
```diff
@@ -0,0 +1,2 @@
+See [here](https://github.com/DataBiosphere/data-explorer-indexers/tree/master/dataset_config/platinum_genomes) for the other dataset config files.
+(`ui.json` is here because it is only used by this repo.)
```
alahwa: I don't think we need to justify why data explorer-specific config is in the data explorer repo

api/dataset_config/platinum_genomes/README.md
```diff
@@ -0,0 +1,2 @@
+See [here](https://github.com/DataBiosphere/data-explorer-indexers/tree/master/dataset_config/platinum_genomes) for the other dataset config files.
```
alahwa: This README seems a bit redundant of the instructions.

README.md
```diff
@@ -20,11 +20,11 @@ https://github.com/DataBiosphere/data-explorer-indexers, or any other indexer.
 * Create `api/dataset_config/<my dataset>`
   * If you used https://github.com/DataBiosphere/data-explorer-indexers, copy
     the config directory from there.
-  * If you used your own indexer, the config files must follow
-    this format ([see examples](https://github.com/DataBiosphere/data-explorer-indexers/tree/master/bigquery/dataset_config/platinum_genomes)). Specifically:
-    * `dataset.json` has a `name` field.
-    * `facet_fields.csv` has the `project_id, dataset_id, table_name, readable_field_name`
-    columns filled out.
+  * Copy and fill out [ui.json](https://github.com/DataBiosphere/data-explorer/tree/master/api/dataset_config/template/ui.json).
```
alahwa: Could we document somewhere that ui.json contains the config for specifying which facets you want to see, and you should fill it out with whatever you're interested in looking at? I don't think that's explicitly mentioned anywhere but isn't obvious to me walking through the steps and filling out config.

README.md
```diff
@@ -20,11 +20,11 @@ https://github.com/DataBiosphere/data-explorer-indexers, or any other indexer.
...
+    (`ui.json` is not in `data-explorer-indexers` repo.)
+  * If you used your own indexer, copy the config files from [here](https://github.com/DataBiosphere/data-explorer-indexers/tree/master/dataset_config/template)
+    and [here](https://github.com/DataBiosphere/data-explorer/tree/master/api/dataset_config/template).
```
alahwa: This link's also broken for me!

README.md
alahwa: This link's throwing a 404

.gitignore
```diff
@@ -1,8 +1,20 @@
...
 
+!api/dataset_config/template
+
+# Most Platinum Genomes config files are in data-explorer-indexers repo. 
+# (ui.json is in this repo since it's only used by this repo.)
+# Only store ui.json; eg we don't want to maintain 2 copies of dataset.json.
+# Negating a previously excluded directory is tricky. See
+# https://stackoverflow.com/a/18702826
+# Note that above must have api/dataset_config/** and not api/dataset_config/*
+!api/dataset_config/platinum_genomes/
```
alahwa: Shouldn't this directory be ignored excluding the two files below? So it should be:

```
api/dataset_config/**
!api/dataset_config/platinum_genomes/ui.json
!api/dataset_config/platinum_genomes/README.md
```

api/data_explorer/__main__.py
```diff
@@ -65,6 +65,21 @@
 app.add_api('swagger.yaml', base_path=args.path_prefix)
```
alahwa: The previous design worked out well because it didn't need the user to specify what the table names were, it just got that information from the list of facets. That had the added benefit of only getting tables that were actually present in the Data Explorer UI.

By de-coupling the two, we're introducing the possibility of exporting tables that don't exist in the Data Explorer facets or forgetting to export tables that do, and putting the burden of correctness on the user. Furthermore, we don't have a clear path towards a smarter export function that links each BQ facet with its data source.

I think it's worth investing into a solution that links facets to their data sources on the back end, which we can use on the front end to do stuff like export to Saturn. What do you think?

api/dataset_config/test_dataset/ui.json
```diff
@@ -0,0 +1,30 @@
+// Configuration for Data Explorer UI
+{
+  "facets": [
+    {
+      "elasticsearch_field_name": "Age",
```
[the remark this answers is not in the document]
melissachang: Yes, will be different in the future

api/data_explorer/__main__.py
melissachang: Good point. I spent some time thinking about this. I think a saved cohort query would look something like:
```
SELECT DISTINCT t1.participant_id
FROM
  (SELECT participant_id
   FROM `project1.dataset1.table1`
   WHERE foo = "FOO") t1
INNER JOIN
  (SELECT participant_id
   FROM `project2.dataset2.table2`
   WHERE bar = "BAR") t2
ON
  t1.participant_id = t2.participant_id
```
Right now, ES index only contains BQ column name ("foo"):
```
{
    foo: "FOO",
}
```
I can change the index to include project/dataset/table:
```
{
    project1.dataset1.table1.foo: "FOO",
}
```
That should give us enough information to construct the query.

This also fixes a problem with the current implementation, where if there 2 tables with field "foo", one would overwrite the other.

Can I do this in a separate PR? I'd have to change the indexer repo first.

melissachang: Or, can you work on this when you get to saving cohort queries?

Just to clarify something which I'm sure you already aware of -- First, just export the whole BQ table with no selection.

.gitignore
alahwa: Ah okay, the `!api/dataset_config/platinum_genomes/` is kind of unintuitive but as long as I can't check in `platinum_genomes/random_file` I'm ok with this

README.md
melissachang: I added a line to `ui.json`. By convention, comments go in the .json files. This way, they'll always see the comment; as opposed to only seeing the comment if they read the README first.

api/data_explorer/__main__.py
melissachang: Let's brainstorm at the next sync. This doesn't affect this PR. The goal of this PR is just to read from `ui.json`, which is unrelated to GCS work. If this needs to change for GCS work, I can change it later.

api/dataset_config/platinum_genomes/README.md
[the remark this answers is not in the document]
melissachang: This comment is for people looking at this directory directly (say, in Github). I put it for the relevant directories. If I were new to Data Explorer and I saw this directory, I would wonder why there's only one config file, when I know there are others. I don't see the harm of duplicating this comment.

api/dataset_config/test_dataset/ui.json
```diff
@@ -0,0 +1,30 @@
+// Configuration for Data Explorer UI
```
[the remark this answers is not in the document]
melissachang: Could be many things. If you want to have a textbox search with autocomplete, if you want certain visualizations (like a map), etc. We have may a setting for export button, for datasets that aren't integrated with Saturn.

PR #95, opened 2018-07-09
Friendly error message for wrong Elasticsearch field name

melissachang:
Before:
```
apise_1            | Traceback (most recent call last):
apise_1            |   File "/usr/local/lib/python2.7/site-packages/flask/app.py
", line 1982, in wsgi_app
apise_1            |     response = self.full_dispatch_request()
apise_1            |   File "/usr/local/lib/python2.7/site-packages/flask/app.py
", line 1607, in full_dispatch_request
apise_1            |     self.try_trigger_before_first_request_functions()
apise_1            |   File "/usr/local/lib/python2.7/site-packages/flask/app.py
", line 1654, in try_trigger_before_first_request_functions
apise_1            |     func()
apise_1            |   File "/app/data_explorer/__main__.py", line 180, in init
apise_1            |     app.app.config['ELASTICSEARCH_FACETS'] = _get_facets()
apise_1            |   File "/app/data_explorer/__main__.py", line 149, in _get_
facets
apise_1            |     field_type = mapping['type']['properties'][field_name][
'type']
apise_1            | KeyError: u'Maternal_grandmother_race_ethnicity'
```
After:
```
apise_1            | ValueError: Field Maternal_grandmother_race_ethnicity not found in Elasticsearch index personal_genome_project
```
Also include change to `ui.json` that I forgot to push in previous PR.


api/data_explorer/__main__.py
```diff
@@ -146,6 +146,9 @@ def _get_facets():
 
     for facet_config in facets_config:
         field_name = facet_config['elasticsearch_field_name']
+        if not field_name in mapping['type']['properties']:
+            raise ValueError('Field %s not found in Elasticsearch index %s' %
```
alahwa: nit: `elasticsearch_field_name %s not found...`

api/data_explorer/__main__.py
```diff
@@ -146,6 +146,9 @@ def _get_facets():
 
     for facet_config in facets_config:
         field_name = facet_config['elasticsearch_field_name']
+        if not field_name in mapping['type']['properties']:
```
alahwa: Let's keep an eye on how sensitive this is, we might consider making this a warning instead of an error in the future.

melissachang: Why would this be a warning? Most likely they made a typo in `ui.json`.

api/data_explorer/__main__.py
melissachang: I think mine is more readable. I think it's obvious we're talking about an Elasticsearch field.

api/data_explorer/__main__.py
alahwa: A warning would be more lenient regarding stuff like changing elasticsearch data, configs for multiple datasets, etc. I don't think we should make it a warning now, this was more of a comment than a suggestion.

api/data_explorer/__main__.py
alahwa: For me, if I run docker-compose up and get this error, it's not obvious that "Field" means "one of the elasticsearch_field_names in ui.json config". Being specific reduces obstacles to debugging

api/data_explorer/__main__.py
melissachang: Yes, we can discuss in the future. But in general, I think error is better. For example, if Elasticsearch data changed and a field is no longer available, it's better to fail and let user know, so they can update to the right field.

api/data_explorer/__main__.py

PR #96, opened 2018-07-10
Disable bootstrap checks for non-prod deployments

melissachang:


alahwa:
One thing to consider is that the purpose of this bootstrap check is to ensure there's enough vm memory to run Elasticsearch. On CircleCI, the vm.max_map_count is set to ~60,000 which is high enough for most purposes (including ours) but we don't know anything about what this value will be on a user's computer. So while we get more simplicity, it could potentially cause out of memory exceptions for the user.

melissachang:
Ok, let's try it and see if anyone gets OOM.

PR #97, opened 2018-07-11
Add optional facet descriptions

melissachang:
description

NHS data is complicated; its facets need descriptions.

### DIFF
--- a/api/data_explorer/__main__.py
+++ b/api/data_explorer/__main__.py
@@ -121,14 +121,30 @@ def _get_table_names():
     return table_names
 
 
-def _get_facets():
-    """Gets facets.
+def _get_ui_facets():
+    """Returns a dict from UI facet name to UI facet description.
 
-    Returns:
-        A list of tuples, where each tuple corresponds to a facet. Each tuple
-        contains: UI facet name, UI facet description, and Elasticsearch facet
-        object.
+    If there is no description for a facet, the value is None.
     """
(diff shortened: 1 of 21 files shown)


api/data_explorer/__main__.py
```diff
@@ -122,10 +122,12 @@ def _get_table_names():
 
 
 def _get_facets():
-    """Gets Elasticsearch facets.
+    """Gets facets.
 
     Returns:
-        A dict from UI facet name to Elasticsearch facet object
+        A list of tuples, where each tuple corresponds to a facet. Each tuple
```
alahwa: I could be missing something here, but are there any use cases that require us to have both the UI facet description and the es facet object? If not, we could leave `app.config['ELASTICSEARCH_FACETS']` as a dict of facet names to es objects and separately have an `app.config['UI_FACETS']` that's either a list or a dict of `Facet` objects.

That would let us keep the ES dict where it's useful in the controller and search instead of having to re-construct it and search through lists while also making these changes extensible if we ever want to expand the API definition of a UI Facet.

api/data_explorer/test/test_facets_controller.py
```diff
@@ -35,6 +35,9 @@ def create_app(self):
         app = super(TestFacetsController, self).create_app()
         app.config.update({
             'INDEX_NAME': 'index_name',
+            'UI_FACETS': {
+                'Region': None
```
alahwa: Could we set to something so we can test it works?

melissachang: Done

PR #100, opened 2018-07-13
Support expanded Elasticsearch BigQuery field names

melissachang:
After https://github.com/DataBiosphere/data-explorer-indexers/pull/44, Elasticsearch index contains field names "google.com:biggene.platinum_genomes.sample_info.gender" instead of "gender". In order to support this, I had to change the way I retrieved field names from Elasticsearch.

Tested:
- Local default dataset
- Local Platinum Genomes

### DIFF
--- a/api/data_explorer/__main__.py
+++ b/api/data_explorer/__main__.py
@@ -111,7 +111,6 @@ def _get_dataset_name():
 
 def _get_table_names():
     """Gets an alphabetically ordered list of table names from facet_fields.csv.
-
     Table names are fully qualified: <project id>:<dataset id>:<table name>
     """
     config_path = os.path.join(app.app.config['DATASET_CONFIG_DIR'],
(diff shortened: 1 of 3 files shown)


melissachang:
In this PR, it doesn't to this repo whether a field is BQ or not. It will matter later, when we implement selected cohort export, but that can be address later.

alahwa:
Even without export, what if we have GCS fields that also satisfy `split('.').length == 3`? Won't that cause issues?

melissachang:
FYI I don't see an option to the old API call, _mapping/type, to not nest. [Documentation here.](https://www.elastic.co/guide/en/elasticsearch/reference/6.3/indices-get-mapping.html#_multiple_indices_and_types)

alahwa:
I did a quick look into this and couldn't find anything easily available either, but since this helps us integrate with Orange box anyway I don't think it matters much.

PR #101, opened 2018-07-16
Update packages

alahwa:
Some of our libraries were deprecated/had known security vulnerabilities/missing peer dependencies. Updating packages to clean up

### DIFF
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -5,36 +5,164 @@
   "requires": true,
   "dependencies": {
     "@babel/code-frame": {
-      "version": "7.0.0-beta.44",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.0.0-beta.44.tgz",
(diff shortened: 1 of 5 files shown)


melissachang:
How much work would it be to migrate all components to material-ui 1.x? If it's less than say a day's worth of work, can you do it in this PR?

Also, can we use one 1.x version, rather than different versions for @material-ui/core and @material-ui/icons?

FYI @karamalhotra

alahwa:
Some of the components are a bit different between 0.x and 1.x as this is a complete re-implementation from the ground up. For example, I tried to convert List/ListItem to 1.x but the spacing looked big and clunky. Migrating will be a process since not all features from 0.x are available in 1.x yet and already 0.x things are starting to become deprecated. But I think it's still worth having 1.x available since that's where all of the new feature work is being done. Eventually I hope to convert everything to 1.x and remove our 0.x dependency, but the library's not fully there yet.

As far as versions, core and icons are maintained separately and have different version numbers. 1.4 and 1.1 are the current versions, respectively.

melissachang:
package.json doesn't support comments, so nothing here:

FYI @karamalhotra -- For new work we're going to try to use material-ui v1. Import from "material-ui/core" instead of "material-ui". It will take a while to migrate everything, so some old stuff is still using 0.x.

PR #102, opened 2018-07-17
Add ExportFab back in

alahwa:


melissachang:
I get this error when I run `docker-compose up --build` locally:
```
ui_1               | Failed to compile.
ui_1               | 
ui_1               | ./src/components/ExportFab.js
ui_1               | Module not found: Can't resolve '@material-ui/core/Button' in '/ui/src/components'
```
I'm not sure how CircleCI passed -- maybe something to do with node_modules caching?

Can you try to repro this error? Things you can try:
- fresh git clone
- deleting containers on your computer

alahwa:
I ran into the same issue, it's actually a really annoying bug with docker's image caching forcing us to restart when we add new packages. I don't know why even `--build` and `--force-recreate` don't do the trick here. Try deleting your existing data explorer images (I just wiped all of my cached images/volumes/containers via `docker system prune`) and re-building.

melissachang:
Can you please share any links that you found in your investigation?

alahwa:
https://github.com/docker/compose/issues/4337

melissachang:
Strange. I did `docker system prune` and things worked for me also.

The problem was with `npm` itself. When I saw the problem, if I `docker exec` into container and run `npm install`, there is no `node_modules/@material-ui/core/Button`. If we keep seeing npm problems, we might consider switching to yarn.

alahwa:
I'm still not sure whether that's an npm problem or not. It's possible the package.json and package-lock.json being copied into the docker container at runtime are coming from a stale cache and not the updated versions, which would be a docker problem. Still, agreed we should keep an eye out for this.

melissachang:
package.json/package-lock.json were correct, it's an npm problem.

PR #106, opened 2018-07-23
Have API server wait until Elasticsearch healthy

melissachang:
With most datasets, Elasticsearch comes up quickly. It just so happens that when I opened http://localhost:4400/, Elasticsearch was ready.

NHS index has > 120k documents, so Elasticsearch takes longer to come up. So I noticed an error when I tried to open http://localhost:4400/ before Elasticsearch was up.

### DIFF
--- a/api/data_explorer/__main__.py
+++ b/api/data_explorer/__main__.py
@@ -67,6 +67,22 @@ app.app.json_encoder = JSONEncoder
 app.add_api('swagger.yaml', base_path=args.path_prefix)
 
 
+def _wait_elasticsearch_healthy():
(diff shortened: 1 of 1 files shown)


api/data_explorer/__main__.py
```diff
@@ -189,6 +206,8 @@ def _get_field_type(es, field_name):
 def _get_es_facets():
     """Returns a dict from UI facet name to Elasticsearch facet object."""
     es = Elasticsearch(app.app.config['ELASTICSEARCH_URL'])
+    _wait_elasticsearch_healthy(es)
```
alahwa: I'd probably call this in init() rather than here

PR #107, opened 2018-07-23
Dynamically determine bucket size for histograms

melissachang:
This is for NHS. Note that I don't want to put too much effort into this, because all this will go away after https://github.com/elastic/elasticsearch/issues/31828 is fixed.

### DIFF
--- a/api/data_explorer/__main__.py
+++ b/api/data_explorer/__main__.py
@@ -232,7 +232,6 @@ def _get_es_facets():
             # TODO: When https://github.com/elastic/elasticsearch/issues/31828
             # is fixed, use AutoHistogramFacet instead. Will no longer need 2
             # steps.
-            # Step 1: Get max value
             response = Search(
                 using=es, index=app.app.config['INDEX_NAME']
             ).aggs.metric(
--- a/api/data_explorer/__main__.py
+++ b/api/data_explorer/__main__.py
@@ -205,6 +205,26 @@ def _get_field_type(es, field_name):
         field_name]['mapping'][last_part]['type']
 
 
+def _get_max_field_value(es, field_name):
+    response = Search(
+        using=es, index=app.app.config['INDEX_NAME']
+    ).aggs.metric(
+        'max',
+        'max',
+        # Don't execute query; we only care about aggregations. See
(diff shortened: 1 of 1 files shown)


api/data_explorer/__main__.py
```diff
@@ -14,6 +14,7 @@
 from elasticsearch_dsl import HistogramFacet
```
alahwa: I think this is fine as a patch for some of the data we currently have, however it won't address numbers much larger than 100, and won't take into account the distribution of the data.

api/data_explorer/__main__.py
```diff
@@ -206,13 +207,31 @@ def _get_es_facets():
...
-            # ranges easy to read (10-19,20-29 instead of 10-17,18-25).
+            # Creating this facet is a two-step process.
+            # 1) Get max value
+            # 2) Based on max value, determine bucket size. Create
+            #    HistogramFacet with this bucket size.
+            # TODO: When https://github.com/elastic/elasticsearch/issues/31828
+            # is fixed, use AutoHistogramFacet instead. Will no longer need 2
+            # steps.
+            # Step 1: Get max value
+            response = Search(
```
alahwa: Could you write a test for this?

melissachang: If it were easy to test, I would. But it's not easy, and this is throw-away code, so I'd rather not add a test...

api/data_explorer/__main__.py
```diff
@@ -224,13 +225,30 @@ def _get_es_facets():
             facets[ui_facet_name] = TermsFacet(field=field_name + '.keyword')
         else:
             # Assume numeric type.
-            # TODO: Handle other types.
-            # TODO: Automatically figure out bucket intervals. Unfortunately
-            # Elasticsearch won't do this for us
-            # (https://github.com/elastic/elasticsearch/issues/9572). Make the
-            # ranges easy to read (10-19,20-29 instead of 10-17,18-25).
+            # Creating this facet is a two-step process.
```
alahwa: Since this function is getting pretty long and has substeps and sub-documentation, split out to helper methods `_get_max_value` and `_get_bucket_size`?

PR #113, opened 2018-07-26
Add exportUrl endpoint

melissachang:


api/api.yaml
```diff
@@ -47,6 +47,24 @@ paths:
       tags:
         # Put in facets_controller.py instead of default_controller.py.
         - Facets
+  /exportUrl:
+    # This is post because this is not idempotent; a GCS file is created.
+    post:
+      description: >
+        Gets a signed URL to a GCS zip file of JSON files. The JSON files
```
alahwa: Gets > "Creates and returns"

api/api.yaml
```diff
@@ -47,6 +47,24 @@ paths:
...
+        represent entites to be exported to a Saturn workspace.
```
alahwa: typo: entites

alahwa: I'd elaborate on "represent entities" to mention that specifically, the JSON contains CreateEntity requests as defined by the Saturn API.

api/api.yaml
```diff
@@ -47,6 +47,24 @@ paths:
...
+        https://bvdp-saturn-prod.appspot.com/#import-data will be called with
```
alahwa: The API documentation should define what the API does instead of speculating what the client should or shouldn't do with the returned results. Lines 56-59 are true, but that documentation probably belongs on the UI side.

api/api.yaml
melissachang: Done

api/api.yaml
melissachang: Done

melissachang: Done. It's not "Saturn API" -- I specifically mention https://rawls.dsde-prod.broadinstitute.org/#!/entities/create_entity to clarify the exact API.

api/api.yaml
melissachang: Done

PR #118, opened 2018-08-08
Check for deploy.json and export bucket

melissachang:
This is initialization for Export to Saturn feature. Next PR will write JSON entities to a
GCS file.

If deploy.json or export bucket is missing, a warning is printed in API server logs. Data Explorer UI still loads, in case people want to play around with Data Explorer UI. Then if they click the Export button, an error message is shown.

Tested locally:
- deploy.json missing - When I click export button, I see error
- bucket missing - When I click export button, I see error
- deploy.json and bucket exist - When I click export button, I'm redirected to Saturn. This confirms local auth works.

Test on GCP
- deploy.json - N/A, deploy/deploy-api.sh won't work if deploy.json is missing
- bucket missing - When I click export button, I see error
- deploy.json and bucket exist. When I click on export button, I'm redirected to Saturn. This confirms GCP auth works. See http://melchang-test-project-3.appspot.com/

CC @malathi13 FYI


api/data_explorer/__main__.py
```diff
@@ -287,6 +288,33 @@ def _get_table_names():
...
+                               'deploy.json')
+    if not os.path.isfile(config_path):
+        app.app.logger.warning(
+            'deploy.json not found. Export to Saturn feature will not work. '
+            'See https://github.com/DataBiosphere/data-explorer#one-time-setup-for-export-to-saturn-feature'
+        )
+        return ''
+
+    project = _parse_json_file(config_path)['project_id']
+    bucket = project + '-export'
```
bfcrampton: tiny nit: maybe '-export-saturn' to make it more specific? I could potentially see someone already having a bucket named `<project>-export`

api/data_explorer/controllers/export_url_controller.py
```diff
@@ -1,13 +1,29 @@
...
 from flask import current_app
+from werkzeug.exceptions import BadRequest
 
 from data_explorer.models.export_url_response import ExportUrlResponse  # noqa: E501
 
 
 def export_url_post():  # noqa: E501
+    config_path = os.path.join(current_app.config['DATASET_CONFIG_DIR'],
+                               'deploy.json')
+    if not os.path.isfile(config_path):
```
bfcrampton: Not sure I understand why this is being checked again. It's already done on startup and if `deploy.json` doesn't exist `current_app.config['EXPORT_URL_GCS_BUCKET']` is set to `''`, so shouldn't the check below suffice?

api/data_explorer/__main__.py
melissachang: So we recommend creating a project just for deploying data explorer (see ["We recommend creating a project"](https://github.com/DataBiosphere/data-explorer-indexers/tree/master/bigquery/deploy), so `<project>-export` probably won't exist.

Also, Saturn is a code name. I mention "Saturn" in comments and READMEs, but when we get a real name, I'll replace "Saturn".

api/data_explorer/controllers/export_url_controller.py
melissachang: This will give a more specific error message. If `deploy.json` doesn't exist, the error "deploy.json not found" is a little more helpful than "Bucket not found".

bfcrampton: Gotcha yeah that makes sense.

PR #124, opened 2018-08-14
Add instructions for creating export url bucket

melissachang:


deploy/create-export-url-bucket.sh
```diff
@@ -0,0 +1,33 @@
+  #!/bin/bash -xb
```
bfcrampton: tiny nit: rm leading space?

deploy/create-export-url-bucket.sh
```diff
@@ -0,0 +1,33 @@
...
+#
+# Creates GCS bucket for Export to Saturn feature.
+#
+# See export_url_controller.py for more information.
+#
+# jq, gsutil must be installed before running this script.
```
bfcrampton: optional: could add a check for this and print out an error saying install `jq` or `gsutil`

deploy/create-export-url-bucket.sh
```diff
@@ -0,0 +1,33 @@
...
+if (( $# != 1 ))
+then
+  echo "Usage: deploy/create-export-url-bucket.sh <dataset>"
+  echo "  where <dataset> is the name of a directory in dataset_config/"
+  echo "Run this script from project root"
+  exit 1
+fi
+
+dataset=$1
+project_id=$(jq --raw-output '.project_id' dataset_config/${dataset}/deploy.json)
```
bfcrampton: Would it be easier to just make this a python script so you get things like json parsing for free? The cloud storage python API seems to support all the things you need for bucket management: https://googlecloudplatform.github.io/google-cloud-python/latest/storage/buckets.html

deploy/create-export-url-bucket.sh
melissachang: Done

deploy/create-export-url-bucket.sh
melissachang: If jq and gsutil aren't installed, output is:
```
melchang@melchang:~/data-explorer$ deploy/create-export-url-bucket.sh 1000_genomes
deploy/create-export-url-bucket.sh: line 18: jq: command not found
deploy/create-export-url-bucket.sh: line 21: gsutil: command not found
Creating gs://-export
deploy/create-export-url-bucket.sh: line 27: gsutil: command not found
Setting bucket TTL to 1 day
deploy/create-export-url-bucket.sh: line 30: gsutil: command not found
Setting up CORS
deploy/create-export-url-bucket.sh: line 33: gsutil: command not found
```
I don't think it's worth detecting if they're installed, because the output is already pretty clear that they should be installed.

deploy/create-export-url-bucket.sh
melissachang: Shell was easier to write

PR #126, opened 2018-08-14
Added a dialog to provide cohort name when doing Export.

malathi13:


ui/src/components/ExportFab.js
```diff
@@ -5,16 +5,30 @@ import "./ExportFab.css";
...
   constructor(props) {
     super(props);
-    this.handleClick = this.handleClick.bind(this);
+    this.state = {open: false};
+    this.handleClick = this.handleClick.bind(this);     
+    this.handleCancel = this.handleCancel.bind(this);    
+    this.handleSave = this.handleSave.bind(this);    
   }
 
+  state = {
```
melissachang: I don't think you need this, `state` is initialized in constructor

ui/src/components/ExportFab.js
```diff
@@ -5,16 +5,30 @@ import "./ExportFab.css";
 import CloudUpload from "@material-ui/icons/CloudUpload";
 import Button from "@material-ui/core/Button";
 import Tooltip from "@material-ui/core/Tooltip";
+import TextField from '@material-ui/core/TextField';
```
melissachang: I get an error and warning when I run locally with your branch, does it work for you?

ui/src/components/ExportFab.js
```diff
@@ -5,16 +5,25 @@ import "./ExportFab.css";
...
```
melissachang: Can be a separate PR -- Can you make the dialog only show up if something is selected?

ui/src/components/ExportFab.js
```diff
@@ -25,10 +34,42 @@ class ExportFab extends React.Component {
           <CloudUpload />
         </Button>
       </Tooltip>
+       <div>
+        <Dialog
```
melissachang: Sorry changed my mind, can you add the Cancel button back in? Thanks.

ui/src/components/ExportFab.js
```diff
@@ -5,16 +5,25 @@ import "./ExportFab.css";
...
-    this.handleClick = this.handleClick.bind(this);
+    this.state = {open: false};
+    this.handleClick = this.handleClick.bind(this);     
+    this.handleSave = this.handleSave.bind(this);    
   }
 
   render() {
     return (
+      <div>
       <Tooltip title="Export to Saturn">
```
melissachang: Can you fix the formatting? Eg this line should be indented two spaces compared to previous line. Also, this I think this div should end on line 37 instead of line 63.

malathi13: There should be an outer <div> enclosing all the tags, I got an error without it.

ui/src/components/ExportFab.js
malathi13: Done

PR #128, opened 2018-08-20
Show the cohort dialog only if one or more filters are selected.

malathi13:


ui/src/App.js
```diff
@@ -100,8 +104,10 @@ class App extends Component {
 
     let filterArray = this.filterMapToArray(this.filterMap);
     if (filterArray.length > 0) {
+      this.setState({ filter: filterArray });
```
melissachang: Move this to before if statement (line 106). Delete lines 107, 110

PR #132, opened 2018-08-21
Add e2e test documentation to clear es index

bfcrampton:


README.md
```diff
@@ -140,6 +140,8 @@ End-to-end tests use [Puppeteer](https://github.com/GoogleChrome/puppeteer) and
 To run locally:
 
 ```
+# Ensure the elasticsearch index is clean
+docker-compose down
```
melissachang: How about `curl -XDELETE localhost:9200/_all` after the `docker-compose up`?

`docker-compose down` will remove containers, which means the next `docker-compose up` will have to rebuild from scratch (I think).

PR #129, opened 2018-08-21
Cache 1000 genomes index in JSON file for easier quickstart

bfcrampton:

PR #137, opened 2018-08-27
Fix JSON index dump for 1000 genomes

bfcrampton:
We were not properly setting the document ID, so elasticsearch auto-generated one for us.


bfcrampton:
I actually accidentally checked in two in this PR: https://github.com/DataBiosphere/data-explorer/pull/129/files. 

In this PR I actually fixed it and deleted the one at the project root (which is why there are two listed, but one is 3500 lines deleted).

PR #140, opened 2018-09-04
Add utility for deleting entities in a workspace

melissachang:
FYI @malathi13

### DIFF
--- a/util/delete-entities.py
+++ b/util/delete-entities.py
@@ -3,32 +3,36 @@
 This is useful when working on Export to Saturn feature.
 
 To run:
-- virtualenv ~/virtualenv/firecloud
-- source ~/virtualenv/firecloud/bin/activate
-- pip install firecloud
-- python util/delete-entities.py WORKSPACE_NAMESPACE WORKSPACE_NAME
+
+virtualenv ~/virtualenv/firecloud
+source ~/virtualenv/firecloud/bin/activate
+pip install firecloud
+python util/delete-entities.py WORKSPACE_NAMESPACE WORKSPACE_NAME
 """
 
+import argparse
+import collections
 import sys
 
 from firecloud import fiss
 
 
+def parse_args():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--workspace-namespace', required=True)
+    parser.add_argument('--workspace-name', required=True)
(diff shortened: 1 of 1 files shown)


util/delete-entities.py
```diff
@@ -0,0 +1,51 @@
...
+def main():
+    if len(sys.argv) != 3:
+        raise ValueError(
+            'Usage: python util/delete-entities.py WORKSPACE_NAMESPACE WORKSPACE_NAME'
+        )
+
+    workspace_namespace = sys.argv[1]
+    workspace_name = sys.argv[2]
+
+    args = type('obj', (object, ), {
```
bfcrampton: I think the standard way to do this is by using a [namedtuple](https://docs.python.org/2/library/collections.html#collections.namedtuple) here:
```
Args = collections.namedtuple('Args', ['project', 'workspace'])
args = Args(workspace_namespace, workspace_name)
```

Though this is pretty neat, didn't realize it was possible this way.

util/delete-entities.py
```diff
@@ -0,0 +1,51 @@
...
+
+    # entities are sorted by type: participant, participant_set, sample,
+    # sample_set. FireCloud complains if we delete a participant before deleting
+    # associated participant_set/sample/sample_set. By reversing the list, we
+    # won't have this problem. We will delete: sample_set, sample,
+    # participant_set, participant.
+    for entity in reversed(entities):
+        print('Deleting ' + entity)
+        args = type(
+            'obj', (object, ), {
```
bfcrampton: I would probably just extract the namedtuple out to the top level:

```
Args = collections.namedtuple('Args', ['project', 'workspace', 'yes', 'entity_type', 'entity'])
```

Then above:
```
entities = fiss.entity_list(Args(workspace_namespace, workspace_name))
```

and here:
```
entity_split entity.split('\t')
fiss.entity_delete(Args(workspace_namespace, workspace_name, True, entity_split[0], entity_split[1]))
```

Could also use keyword arguments if you think that makes it clearer.

util/delete-entities.py
melissachang: Done

util/delete-entities.py
melissachang: Done

PR #146, opened 2018-09-12
Add /fields endpoint; add react-select component to UI

malathi13:
Implements the /fields api method to return a fixed set of fields for now.
Added the react-select component to the App to enable autocomplete for field exploration.


melissachang:
Malathi and I saw this error in UI server:
```
ui_1               | Failed to compile.
ui_1               | 
ui_1               | ./src/App.js
ui_1               | Module not found: Can't resolve 'react-select' in '/ui/src'
ui_1               | Compiling...
ui_1               | Failed to compile.
ui_1               | 
ui_1               | ./src/App.js
ui_1               | Module not found: Can't resolve 'react-select' in '/ui/src'
```

The problem has [something to do with anonymous volumes](https://github.com/docker/compose/issues/6183#issuecomment-420836250). Workaround is to pass `-V` to `docker-compose up`'. `-V` only needs to be passed once; subsequent `docker-compose up`'s don't need it.

ui/src/App.js
```diff
@@ -3,9 +3,9 @@ import { ApiClient, DatasetApi, FacetsApi } from "data_explorer_service";
 import ExportFab from "./components/ExportFab";
 import FacetsGrid from "./components/facets/FacetsGrid";
 import Header from "./components/Header";
-import FieldSearchTextField from "./components/FieldSearchTextField";
 
 import React, { Component } from "react";
+import Select from "react-select";
```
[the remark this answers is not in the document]
melissachang: I'd prefer creating the component in this PR. It's easy to move fields and selectedOption around if we need to. It's strange to put this in App component.

For now, you can omit selectedOption if you want. Just pass fields into the props of FacetSearch component.

FYI - Interaction would similar to [how facet clicks are handled](https://github.com/DataBiosphere/data-explorer/blob/master/ui/src/components/facets/FacetCard.js#L101)

api/api.yaml
```diff
@@ -134,3 +145,25 @@ definitions:
...
+    properties:
+      fields:
+        type: array
+        items:
+          $ref: '#/definitions/Field'
+  Field:
+    description: >
+      A field. For example, a BigQuery Gender column could include the field
+      name "Gender", description "Sex at birth" and elasticsearch field name
+       "project_id.dataset_id.table_name.Gender".
```
melissachang: nit delete extra space in front

api/api.yaml
```diff
@@ -134,3 +145,25 @@ definitions:
...
+    properties:
+      name:
+        type: string
+        description: Field name, for example, "Gender".
+      description:
+        type: string
+        description: Optional field description.
+      elasticsearch_name:
+        type: string
+        description: The Elasticsearch name.
```
melissachang: The Elasticsearch field name.

PR #147, opened 2018-09-12
Fix auto binning by using range

bfcrampton:


api/data_explorer/__main__.py
```diff
@@ -180,47 +180,53 @@ def _get_field_type(es, field_name):
         field_name]['mapping'][last_part]['type']
 
 
-def _get_max_field_value(es, field_name):
+def _get_field_value_range(es, field_name):
```
melissachang: In retrospect, my naming is confusing. Here's what I suggest.

_get_field_value_range() -> _get_field_range()
field_value_range -> field_range
_get_interval() -> _get_bucket_interval()

bfcrampton: Done.

PR #150, opened 2018-09-19
Improve README "One-time setup for Export to Saturn feature"

melissachang:
CC @mbookman

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,12 +112,18 @@ for your dataset.
 `deploy.json` will still need to be filled out. A temporary file will be written to a
 GCS bucket in the project in `deploy.json`, even for local deployment of Data
 Explorer. Choose a project where you have at least Project Editor permissions.
-* `deploy/create-export-url-bucket.sh DATASET`, where `DATASET` is the name of
+* Create export bucket. This only needs to be done once per deploy project.  
+`deploy/create-export-url-bucket.sh DATASET`, where `DATASET` is the name of
 the directory in `dataset_config`.
-* The Export to Saturn feature creates a signed URL, which requires a service
-account private key. Follow these instructions (["Generate a new private key"](https://cloud.google.com/storage/docs/access-control/create-signed-urls-program#signing-language))
-to create the private key. Select the App Engine default service account.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -112,12 +112,18 @@ for your dataset.
 `deploy.json` will still need to be filled out. A temporary file will be written to a
 GCS bucket in the project in `deploy.json`, even for local deployment of Data
 Explorer. Choose a project where you have at least Project Editor permissions.
-* `deploy/create-export-url-bucket.sh DATASET`, where `DATASET` is the name of
+* Create export bucket. This only needs to be done once per deploy project.  
+`deploy/create-export-url-bucket.sh DATASET`, where `DATASET` is the name of
```
bfcrampton: Nit maybe: "Run `deploy/create-export-url-bucket.sh DATASET` from the root of the repo, where ..."

README.md
```diff
@@ -112,12 +112,18 @@ for your dataset.
...
 the directory in `dataset_config`.
-* The Export to Saturn feature creates a signed URL, which requires a service
-account private key. Follow these instructions (["Generate a new private key"](https://cloud.google.com/storage/docs/access-control/create-signed-urls-program#signing-language))
-to create the private key. Select the App Engine default service account.
-Store the key in `dataset_config/DATASET/private-key.json`.
+* The Export to Saturn feature requires a service account private key. Follow
+these instructions to download a key. This needs to be done once per person
+per dataset. If three people run Data Explorer
+for dataset D, all three need to download a key for dataset D.
```
bfcrampton: Could this be a little confusing here in the meaning of the word "dataset"? People reading this may think the service account key has to be associated with a particular BigQuery dataset, when it actually just has to be linked to the deployment project. Would replacing `dataset` with `deployment` work in this section? Or maybe `dataset deployment`? 

I'd also just say "for a particular deployment" rather than "dataset D"

README.md

README.md
melissachang: Done. I went with "deploy project" instead of deployment. Technically, someone could run Data Explorer locally and still need a key. (I associate "deployment' with "GCP deployment", not "local deployment".)

PR #151, opened 2018-09-19
Get fields from elastic search index

malathi13:


melissachang:
For the load_index/dump_index changes, let's use the bryancrampton tables. It's ok if fields index doesn't have descriptions.

ui/src/App.js
```diff
@@ -47,6 +48,12 @@ class App extends Component {
       } else {
         this.setState({
           fields: data.fields.map(field => {
+            if (field.description == null || field.description === "") {
```
melissachang: longchip

The chip can be long, see image above. We should show "field name - description" in the drop-down, but only field name in the chip. Can be a separate PR. I [asked about this on StackOverflow](https://stackoverflow.com/questions/52433569/react-select-show-different-text-label-for-drop-down-and-control).

PR #155, opened 2018-09-24
Minor README changes

melissachang:


README.md
```diff
@@ -46,9 +45,33 @@ Run local Data Explorer with the 1000 Genomes dataset:
...
-For local development, an nginx reverse proxy is used to get around CORS.
+The basic flow:
 
-![Architecture overview](https://i.imgur.com/IZLbPx9.png)
+1. Index dataset into Elasticsearch using an indexer from https://github.com/DataBiosphere/data-explorer-indexers
+2. Run the servers in this repo to display Data Explorer UI
+
+GCP deployment:
+
+![GCP deployment architecture](https://i.imgur.com/mCjDv47.png)
```
bfcrampton: This diagram doesn't seem right, I think you may have just mixed up the order? It seems to me like API server should be in the middle (it talks to elasticsearch, not the UI server as shown).

README.md
```diff
@@ -46,9 +45,33 @@ Run local Data Explorer with the 1000 Genomes dataset:
...
+
+For local development, an nginx reverse proxy is used to get around CORS:
+
+![Local deployment architecture](https://i.imgur.com/IZLbPx9.png)
+
+## Want to try out Data Explorer for your dataset?
+
+Here's one possible flow.
+
+- [Run local Data Explorer with public 1000 Genomes dataset.](https://github.com/DataBiosphere/data-explorer#quickstart)
```
bfcrampton: nit: I'd add a note that it's using a cache of the dataset to distinguish it from the next one and make it clear why it doesn't require indexing. Maybe "... with a JSON-cached public 1000 ..."?

README.md

README.md

PR #157, opened 2018-09-25
Read from config files at server startup instead of after first request

melissachang:
Also improve logging of app.app.config:
- Print each key-value pair on a separate line
- Alphabetize keys


api/data_explorer/__main__.py
```diff
@@ -398,24 +398,24 @@ def _process_export_url():
...
     _process_bigquery()
     _process_export_url()
 
-    app.app.logger.info('app.app.config: %s' % app.app.config)
+    app.app.logger.info('app.app.config:')
+    for key in sorted(app.app.config.keys()):
+        app.app.logger.info('    %s: %s' % (key, app.app.config[key]))
+
 
+init()
```
bfcrampton: Why wasn't this needed before?

melissachang:
Reason for this PR:
- Eventually we'll want BQ table descriptions. I figure we should fetch in __main__, rather than on every request. I also figure this would take at least a second, so better to initialize before first request
- I do notice that on app engine, sometimes the first request hangs. I think this PR will fix that. Note that App Engine Flex VMs are ["restarted on a weekly basis"](https://cloud.google.com/appengine/docs/flexible/python/how-instances-are-managed), so the slowness is more than just once per deployment

For local deployment, this does seem to make first page load snappier.

PR #164, opened 2018-09-27
Support adding facets with field search

melissachang:
Malathi, can you please manually test? Thanks.

Differences from your branch:

I changed the react-select label to only be "field name" instead of "field name - description". With the latter, deleting a chip didn't work. this.filterMap key was "Avuncular", but label was "Avuncular - sample ids for any avuncular relationships". (This worked for Malathi because her cached index didn't have field descriptions.)

I pass `extraFacetsOptions` into FieldSearch/Select components as a prop.

In React, state is something a component has to actively manage. Prop is a read-only property that's passed in from parent.

Old PR: Both App and React select components had extraFacetsOptions as state. This means our code has to make sure they're in sync. I'm guessing this would have led to subtle bugs in the future. There may even have been subtle bugs with the PR.

This PR: App component has extraFacetsOption as state. App component passes extraFacetsOptions to React select component via prop. So our code only has to worry about maintaining one extraFacetsOptions.

extraFacetsEsFieldNames is not stored in state, since it can be computed from extraFacetsOptions. In React, [state should be minimal](https://reactjs.org/docs/thinking-in-react.html#step-3-identify-the-minimal-but-complete-representation-of-ui-state). Including both in state might not seem bad, but in practice, it leads to subtle bugs over time, because there will be bugs where the two variables aren't in sync. That's why React has this guideline.

### DIFF
--- a/ui/src/components/FieldSearch.js
+++ b/ui/src/components/FieldSearch.js
@@ -3,20 +3,16 @@ import Select from "react-select";
 
 class FieldSearch extends React.Component {
   constructor(props) {
-    const fields = props.fields;
     super(props);
-    this.handleChange = props.handleChange;
-    this.state = {
-      fields: fields
-    };
   }
 
   render() {
     return (
       <Select
         isMulti="true"
-        onChange={this.handleChange}
-        options={this.state.fields}
+        onChange={this.props.handleChange}
+        options={this.props.fields}
+        value={this.props.extraFacetsOptions}
       />
     );
   }
--- a/api/data_explorer/controllers/facets_controller.py
+++ b/api/data_explorer/controllers/facets_controller.py
@@ -1,13 +1,16 @@
 import pprint
+import urllib
+
+from collections import OrderedDict
+from elasticsearch import Elasticsearch
+from elasticsearch_dsl import HistogramFacet
+from flask import current_app
 
 from data_explorer.models.facet import Facet
 from data_explorer.models.facet_value import FacetValue
 from data_explorer.models.facets_response import FacetsResponse
+from data_explorer.util import elasticsearch_util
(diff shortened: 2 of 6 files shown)


melissachang:
Note that exporting SQL doesn't work for extra facets. (I confirmed it doesn't work your branch also.) You can fix in a future PR.

ui/src/components/FieldSearch.js
```diff
@@ -3,20 +3,16 @@ import Select from "react-select";
...
 
   render() {
     return (
       <Select
         isMulti="true"
-        onChange={this.handleChange}
-        options={this.state.fields}
+        onChange={this.props.handleChange}
+        options={this.props.fields}
+        value={this.props.extraFacetsOptions}
```
malathi13: I actually don't find this intuitive. I think the FieldSearch component is the source of truth for storing the selectedOptions. The FieldSearch component passes the selectedOptions to the parent component in the event handler for onChange(). The parent component is storing the selectedOptions in it's state. 
But in this PR, the parent is setting the value of the select component, which seems like it's overwriting the source. IMO, this is also prone to bugs. If the parent component modifies the selectedOptions then the value of the select box will also change.

melissachang: Discussed offline. Yeah, this aspect of Reactive is unintuitive in the beginning. Intuitively, one would expect that FieldSearch would be source of truth for selected options, and the FacetsGrid would also be source of truth for selected options (since it has to display new facets). When user adds a new option, FieldSearch would notify FacetsGrid. In practice, having different components both maintain the same source of truth leads to bugs, because our code might not keep them in sync.

The React way is, react-select callback tells App that selection changed. App tells all children (including FieldSearch) about new options. It is more roundabout than FieldSearch just knowing itself which options are selected, but it leads to simpler code over time.

PR #172, opened 2018-10-08
Change from relative to absolute imports

melissachang:
This will make maintaining .js files easier.


ui/.env
```diff
@@ -0,0 +1,5 @@
+# Note: After changing this file, must restart React app. Reloading Chrome tab
```
malathi13: Will we need to restart the data explorers deployed on GCP?

melissachang: No, only need to restart if we want them to have a particular commit

PR #180, opened 2018-10-12
Support nested facets in Data Explorer UI

malathi13:
image

### DIFF
--- a/api/data_explorer/__main__.py
+++ b/api/data_explorer/__main__.py
@@ -202,6 +202,11 @@ def _process_facets(es):
             'Samples Overview'] = elasticsearch_util.get_samples_overview_facet(
                 es_field_names)
 
+    # Get the list of nested facet paths.
+    nested_facet_paths = elasticsearch_util.get_nested_paths(es)
+    print("nested facet paths")
+    print(nested_facet_paths)
+
     for facet_config in facets_config:
         elasticsearch_field_name = facet_config['elasticsearch_field_name']
         field_type = elasticsearch_util.get_field_type(
@@ -219,7 +224,7 @@ def _process_facets(es):
                 'ui_facet_description']
 
         es_facets[ui_facet_name] = elasticsearch_util.get_elasticsearch_facet(
-            es, elasticsearch_field_name, field_type)
+            es, elasticsearch_field_name, field_type, nested_facet_paths)
 
     # Map from UI facet name to Elasticsearch facet object
     app.app.config['ELASTICSEARCH_FACETS'] = es_facets
@@ -290,9 +295,10 @@ def init():
         _process_facets(es)
         _process_export_url()
 
-        app.app.logger.info('app.app.config:')
-        for key in sorted(app.app.config.keys()):
-            app.app.logger.info('    %s: %s' % (key, app.app.config[key]))
+        app.app.logger.info('    %s: %s' % ('ELASTICSEARCH_FACETS', app.app.config['ELASTICSEARCH_FACETS']))
+        #app.app.logger.info('app.app.config:')
+        #for key in sorted(app.app.config.keys()):
+        #    app.app.logger.info('    %s: %s' % (key, app.app.config[key]))
 
 
 init()
--- a/api/data_explorer/__main__.py
+++ b/api/data_explorer/__main__.py
@@ -295,7 +295,9 @@ def init():
         _process_facets(es)
         _process_export_url()
 
-        app.app.logger.info('    %s: %s' % ('ELASTICSEARCH_FACETS', app.app.config['ELASTICSEARCH_FACETS']))
+        app.app.logger.info(
+            '    %s: %s' % ('ELASTICSEARCH_FACETS',
+                            app.app.config['ELASTICSEARCH_FACETS']))
         #app.app.logger.info('app.app.config:')
         #for key in sorted(app.app.config.keys()):
         #    app.app.logger.info('    %s: %s' % (key, app.app.config[key]))
--- a/api/data_explorer/util/elasticsearch_util.py
+++ b/api/data_explorer/util/elasticsearch_util.py
@@ -193,29 +193,24 @@ def get_field_type(es, field_name):
         field_name]['mapping'][last_part]['type']
 
 
-def get_nested_field_paths(prefix, field_mappings):
+def get_nested_field_paths(prefix, mappings):
     nested_field_paths = []
-    for field_name, field in field_mappings.items():
+    for field_name, field in mappings.items():
+        nested_path = field_name
+        if prefix:
+            nested_path = '%s.%s' % (prefix, field_name)
         if 'type' in field and field['type'] == 'nested':
-            nested_field_paths.append(('%s.%s.' % (prefix, field_name)))
-            nested_field_paths.extend(get_nested_field_paths('%s.%s' % (prefix, field_name), field['properties']))
-    print(nested_field_paths)
+            nested_field_paths.append(nested_path)
+        if 'properties' in field:
+            nested_field_paths.extend(get_nested_field_paths(nested_path, field['properties']))
     return nested_field_paths
 
 
 def get_nested_paths(es):
     nested_paths = []
     mappings = es.indices.get_mapping(index=current_app.config['INDEX_NAME'])
-    project_mappings = mappings[current_app.config['INDEX_NAME']]['mappings']['type']['properties']
-    for project in project_mappings:
-        dataset_mappings = project_mappings[project]['properties']
-        for dataset in dataset_mappings:
-            table_mappings = dataset_mappings[dataset]['properties']
-            for table in table_mappings:
-                field_mappings = table_mappings[table]['properties']
-                print(field_mappings)
-                nested_paths.extend(get_nested_field_paths('%s.%s.%s' % (project, dataset, table), field_mappings))
-    print(nested_paths)
+    nested_paths.extend(get_nested_field_paths('', mappings[
+        current_app.config['INDEX_NAME']]['mappings']['type']['properties']))
     return nested_paths
 
 
@@ -250,7 +245,7 @@ def get_elasticsearch_facet(es, elasticsearch_field_name, field_type, nested_fac
 
 def is_nested_facet(elasticsearch_field_name, nested_facet_paths):
     for path in nested_facet_paths:
-        if elasticsearch_field_name.startswith(path):
+        if elasticsearch_field_name.startswith(path + '.'):
             return path
     return None
 
--- a/api/data_explorer/util/elasticsearch_util.py
+++ b/api/data_explorer/util/elasticsearch_util.py
@@ -139,7 +139,7 @@ def convert_to_index_name(s):
 
 def get_facet_value_dict(filter_arr, es_facets):
     """
-    Parses an array of filters and es facets into a dict of facet_name:[facet_value] 
+    Parses an array of filters and es facets into a dict of facet_name:[facet_value]
     mappings.
     """
     if not filter_arr or filter_arr == [""]:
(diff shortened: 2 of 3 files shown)


api/data_explorer/util/elasticsearch_util.py
```diff
@@ -215,13 +240,20 @@ def get_elasticsearch_facet(es, elasticsearch_field_name, field_type):
             field=elasticsearch_field_name,
             interval=_get_bucket_interval(field_range))
 
-    # Handle sample facets in a special way since they are nested objects.
-    if elasticsearch_field_name.startswith('samples.'):
-        es_facet = NestedFacet('samples', es_facet)
+    nested_path = is_nested_facet(elasticsearch_field_name, nested_facet_paths)
```
bfcrampton: This will need to be done recursively. If you have `fieldA` which is RECORD type and that has inner `fieldB` which is also RECORD type, that will need to be translated into: 
```
NestedFacet('fieldA', NestedFacet('fieldB', es_facet))
```

api/data_explorer/__main__.py
```diff
@@ -202,6 +202,9 @@ def _process_facets(es):
             'Samples Overview'] = elasticsearch_util.get_samples_overview_facet(
                 es_field_names)
 
+    # Get the list of nested facet paths.
+    nested_facet_paths = elasticsearch_util.get_nested_paths(es)
```
melissachang: nested_paths

(There's nothing about facets in nested_paths)

api/data_explorer/__main__.py
```diff
@@ -202,6 +202,9 @@ def _process_facets(es):
             'Samples Overview'] = elasticsearch_util.get_samples_overview_facet(
                 es_field_names)
 
+    # Get the list of nested facet paths.
```
melissachang: opt - Delete comment, it doesn't add anything

api/data_explorer/util/elasticsearch_util.py
```diff
@@ -193,7 +193,32 @@ def get_field_type(es, field_name):
...
+            nested_path = '%s.%s' % (prefix, field_name)
+        if 'type' in field and field['type'] == 'nested':
+            nested_field_paths.append(nested_path)
+        if 'properties' in field:
+            nested_field_paths.extend(
+                get_nested_field_paths(nested_path, field['properties']))
+    return nested_field_paths
+
+
+def get_nested_paths(es):
```
melissachang: Anything that's not obvious or is complicated should have a comment.
```
"""
Returns nested paths, which can be used to created NestedFacet's.

When performing faceted search on nested fields within a document, NestedFacet must be used.
See https://elasticsearch-dsl.readthedocs.io/en/latest/faceted_search.html?highlight=nestedfacet#configuration

The first argument to NestedFacet is a path to the nested field. For example, the 1000 Genomes index has
one nested path: "samples". See https://github.com/DataBiosphere/data-explorer-indexers#main-dataset-index

This method crawls through index mappings and returns all nested paths.
"""
```

api/data_explorer/util/elasticsearch_util.py
```diff
@@ -193,7 +193,32 @@ def get_field_type(es, field_name):
         field_name]['mapping'][last_part]['type']
 
 
-def get_elasticsearch_facet(es, elasticsearch_field_name, field_type):
+def get_nested_field_paths(prefix, mappings):
```
melissachang: _get_nested_paths_inner

Internal methods begin with _

api/data_explorer/util/elasticsearch_util.py
melissachang: Malathi, can you please set up a Data Explorer to manually test this? Either use an existing table or create one.  Please let me know the table so I can test too. Thanks.

Also, I thought you had to traverse nested_facet_paths backwards, for this case.

api/data_explorer/util/elasticsearch_util.py
```diff
@@ -215,13 +240,20 @@ def get_elasticsearch_facet(es, elasticsearch_field_name, field_type):
...
-    if elasticsearch_field_name.startswith('samples.'):
-        es_facet = NestedFacet('samples', es_facet)
+    nested_path = is_nested_facet(elasticsearch_field_name, nested_facet_paths)
+    if nested_path:
+        es_facet = NestedFacet(nested_path, es_facet)
 
     return es_facet
 
 
+def is_nested_facet(elasticsearch_field_name, nested_facet_paths):
```
melissachang: Internal methods begin with _
By convention, methods are defined before they're used. Please move to before get_elasticsearch_facet().
is_nested_facet() sounds like it returns a boolean. How about:
```
def _maybe_get_nested_path
"""
If given field is a nested field, returns nested path. Otherwise returns None.
"""
```

api/data_explorer/util/elasticsearch_util.py
```diff
@@ -215,13 +240,20 @@ def get_elasticsearch_facet(es, elasticsearch_field_name, field_type):
...
+    for path in nested_facet_paths:
+        if elasticsearch_field_name.startswith(path + '.'):
```
melissachang:
```
# Say we have:
# - "samples" is a nested field
# - "samplesType" is not a nested field.
# The '.' makes sure we don't match against "samplesType".
```

api/data_explorer/util/elasticsearch_util.py
```diff
@@ -193,7 +193,43 @@ def get_field_type(es, field_name):
         field_name]['mapping'][last_part]['type']
 
 
-def get_elasticsearch_facet(es, elasticsearch_field_name, field_type):
+def _get_nested_field_paths(prefix, mappings):
```
melissachang: _get_nested_paths_inner?

(It's not clear what the difference between `get_nested_paths()` and `_get_nested_field_paths()` is.)

api/data_explorer/util/elasticsearch_util.py
```diff
@@ -215,13 +251,39 @@ def get_elasticsearch_facet(es, elasticsearch_field_name, field_type):
...
-        es_facet = NestedFacet('samples', es_facet)
-
+    nested_facet = _get_nested_facet(elasticsearch_field_name, es_facet,
+                                     nested_paths)
+    if nested_facet:
+        es_facet = nested_facet
     return es_facet
 
 
+def _get_nested_facet(elasticsearch_field_name, es_facet, nested_paths):
```
melissachang: I think _maybe_get_nested_facet is more clear

api/data_explorer/util/elasticsearch_util.py
```diff
@@ -215,13 +251,39 @@ def get_elasticsearch_facet(es, elasticsearch_field_name, field_type):
...
+    """
+    Returns a NestedFacet for the elasticsearch field, if the field is nested.
```
melissachang: nit Elasticsearch

api/data_explorer/util/elasticsearch_util.py
```diff
@@ -215,13 +251,39 @@ def get_elasticsearch_facet(es, elasticsearch_field_name, field_type):
...
+    """
+    parent = elasticsearch_field_name.rsplit('.', 1)[0]
+    nested_facet = None
+    is_nested = True
```
melissachang: This is confusing, elasticsearch_field_name may not be nested.

Can this method just be:
```
# Say nested_paths is: ["outer", "outer.inner"] and elasticsearch_field_name is "outer.inner.field".
# Traverse nested_paths in reverse order to make sure we use "outer.inner" instead of "outer".
for path in nested_paths in backwards order:
  if elasticsearch_field_name = '.' in path:
    return NestedFacet
```

If it is this simple, it doesn't need to be a separate method -- up to you.

api/data_explorer/util/elasticsearch_util.py
malathi13: Done

melissachang: Not done, did you forget to push?

api/data_explorer/util/elasticsearch_util.py
```diff
@@ -215,13 +251,44 @@ def get_elasticsearch_facet(es, elasticsearch_field_name, field_type):
...
     return es_facet
 
 
+def _get_nested_facet(elasticsearch_field_name, es_facet, nested_paths):
+    """
+    Returns a NestedFacet for the Elasticsearch field, if the field is nested.
+    """
+    parent = elasticsearch_field_name.rsplit('.', 1)[0]
+    nested_facet = None
+    # Traverse up the nesting levels from the leaf field, till we reach the root.
```
melissachang: Can you add somewhere:

Note there can be multiple levels of NestedFacet, eg NestedFacet(outer, NestedFacet(inner, es_facet))

api/data_explorer/util/elasticsearch_util.py
```diff
@@ -215,13 +251,44 @@ def get_elasticsearch_facet(es, elasticsearch_field_name, field_type):
...
-        es_facet = NestedFacet('samples', es_facet)
-
+    nested_facet = _get_nested_facet(elasticsearch_field_name, es_facet,
+                                     nested_paths)
+    if nested_facet:
+        es_facet = nested_facet
     return es_facet
 
 
+def _get_nested_facet(elasticsearch_field_name, es_facet, nested_paths):
```
melissachang: Move to before get_elasticsearch_facet()

api/data_explorer/util/elasticsearch_util.py
```diff
@@ -215,13 +251,44 @@ def get_elasticsearch_facet(es, elasticsearch_field_name, field_type):
...
+    Returns a NestedFacet for the Elasticsearch field, if the field is nested.
+    """
+    parent = elasticsearch_field_name.rsplit('.', 1)[0]
+    nested_facet = None
+    # Traverse up the nesting levels from the leaf field, till we reach the root.
+    # Need to traverse till the root, because the root can be a nested field,
+    # for example "samples". All the sub fields can be non-nested, like
+    # "samples.verily-public-data.human_genome_variants.1000_genomes_sample_info.Main_project_LC_platform"
+    # This field needs to be a NestedFacet because an ancestor("samples") is nested.
+    while parent:
```
melissachang: Here's a simpler version, does this work?
```
def _get_nested_facet(elasticsearch_field_name, es_facet, nested_paths):
    """
    Returns a NestedFacet for the Elasticsearch field, if the field is nested.
    """
    parts = elasticsearch_field_name.rsplit('.', 1)
    # Traverse up the nesting levels from the leaf field, till we reach the root.
    # Need to traverse till the root, because the root can be a nested field,
    # for example "samples". All the sub fields can be non-nested, like
    # "samples.verily-public-data.human_genome_variants.1000_genomes_sample_info.Main_project_LC_platform"
    # This field needs to be a NestedFacet because an ancestor("samples") is nested.
    while len(parts) > 1:
        parent = parts[0]
        if parent in nested_paths:
            es_facet = NestedFacet(parent, es_facet)
        parts = parent.rsplit('.', 1)

    return es_facet
```

api/data_explorer/util/elasticsearch_util.py
malathi13: Oh, didn't see that "maybe" was part of the method name.

api/data_explorer/util/elasticsearch_util.py
malathi13: Yeah that is simpler, Python simplifies array search.

api/data_explorer/util/elasticsearch_util.py
melissachang: nit not fixed

api/data_explorer/util/elasticsearch_util.py
melissachang: Not done.

(It's hard to see unresolved comments in Github UI, I look at the github email)

PR #183, opened 2018-10-16
Delete UI unit tests

melissachang:
It would have been non-trivial to fix them for an upcoming PR. In practice they haven't been useful, and have just been a maintenance burden. For example, the end-to-end tests frequently catch bugs for me, but the UI unit tests haven't caught any. (Nor have the API unit tests.)


ui/package.json
```diff
@@ -22,51 +18,26 @@
     "precommit": "pretty-quick --staged",
     "start": "PORT=4400 react-scripts start",
     "build": "react-scripts build",
-    "test": "jest tests/unit",
-    "update": "jest tests/unit --updateSnapshot",
-    "test:e2e": "jest --config '{\"preset\":\"jest-puppeteer\"}' tests/e2e",
```
malathi13: Should this be added to the line below?

melissachang: I added this on line 41.

.circleci/config.yml
```diff
@@ -31,15 +31,10 @@ jobs:
             pip install yapf
             yapf -dr . --exclude ui/node_modules/
       - run:
-          name: Run the ui linter
+          name: Run the UI linter
           command: |
               cd ui
               npm run lint -- --list-different
```
bfcrampton: Total separate, but we may want to add to the README the command you gave me to auto-fix linting issues?

melissachang: Good idea, done. I also added `--list-different` to package.json and took it out of .circleci/config.yaml.

PR #186, opened 2018-10-17
Key the facets config and filters by elasticsearch field name

malathi13:


melissachang:
I just wanted to document the motivation for this PR.

Prior to this PR, we keyed on UI facet name.

1000 Genomes Data Explorer has a [Population facet](https://github.com/DataBiosphere/data-explorer/blob/master/dataset_config/1000_genomes/ui.json#L15) that points to verily-public-data.human_genome_variants.1000_genomes_participant_info.**Population_Description**.

With the new field search feature, one could add a Population facet that points to verily-public-data.human_genome_variants.1000_genomes_participant_info.**Population**. Now the UI facet name "Population" is ambiguous.

Keying on Elasticsearch field name removes the ambiguity.

api/data_explorer/controllers/export_url_controller.py
```diff
@@ -236,20 +236,18 @@ def _get_range_clause(column, value):
...
-def _get_table_and_clause(facet, value, sample_file_column_fields):
+def _get_table_and_clause(es_field_name, type, value,
+                          sample_file_column_fields):
     """Returns a table name and a single condition of a WHERE clause,
     eg "((age76 >= 20 AND age76 < 30) OR (age76 >= 30 AND age76 < 40))".
     """
     sample_file_type_field = False
-    es_field_name = ''
-    if facet['type'] == 'samples_overview':
+    if type == 'samples_overview':
```
bfcrampton: nit: `type` is [built-in](https://docs.python.org/2/library/functions.html#type) python function and should be avoided as a variable name.

api/data_explorer/__main__.py
```diff
@@ -212,19 +212,18 @@ def _process_facets(es):
...
-            es, elasticsearch_field_name, field_type, nested_paths)
+        ui_facets[elasticsearch_field_name][
+            'facet'] = elasticsearch_util.get_elasticsearch_facet(
+                es, elasticsearch_field_name, field_type, nested_paths)
 
-    # Map from UI facet name to Elasticsearch facet object
-    app.app.config['ELASTICSEARCH_FACETS'] = es_facets
     # Map from UI facet name to dict with Elasticsearch field name,
     # Elasticsearch field type, and optional UI facet description.
     app.app.config['UI_FACETS'] = ui_facets
```
bfcrampton: nit: Might call this `'FACET_INFO'` or something like that now

api/data_explorer/controllers/export_url_controller.py
```diff
@@ -236,20 +236,18 @@ def _get_range_clause(column, value):
 def _get_facet_and_value(filter):
     facets = current_app.config['UI_FACETS']
     split = filter.rsplit('=', 1)
```
bfcrampton: nit: I know this wasn't added, but `filter` is also a built-in python function https://docs.python.org/2/library/functions.html#filter. Maybe `filter_value`?

api/data_explorer/__main__.py
```diff
@@ -196,10 +195,11 @@ def _process_facets(es):
             es_field_names[facet_name] = es_field_name
         ui_facets['Samples Overview'] = {
             'elasticsearch_field_names': es_field_names,
-            'type': 'samples_overview'
+            'type': 'samples_overview',
+            'ui_facet_name': 'Samples Overview'
         }
-        es_facets[
-            'Samples Overview'] = elasticsearch_util.get_samples_overview_facet(
+        ui_facets['Samples Overview'][
```
bfcrampton: tiny nit: maybe `samples_overview` to match the format of the other keys?

api/data_explorer/controllers/export_url_controller.py
```diff
@@ -285,9 +282,9 @@ def _get_filter_query(filters):
 
     table_columns = dict()
     for filter_str in filters:
-        facet, value = _get_facet_and_value(filter_str)
+        es_field_name, facet, value = _get_facet_and_value(filter_str)
```
bfcrampton: nit: I might change `es_field_name` to `facet_name`. In the case of the Samples Overview facet, this isn't actually an ES field name, and we may end up having more `FiltersFacets` in the future, where there isn't a single ES field.

api/data_explorer/controllers/facets_controller.py
```diff
@@ -79,27 +79,27 @@ def facets_get(filter=None, extraFacets=None):  # noqa: E501
...
     #     'Elasticsearch request: %s' % pprint.pformat(search.build_search().to_dict()))
     es_response = search.execute()
     es_response_facets = es_response.facets.to_dict()
     # Uncomment to print Elasticsearch response python object
-    # current_app.logger.info('Elasticsearch response: %s' % pprint.pformat(es_response_facets))
+    #current_app.logger.info(
+    #    'Elasticsearch response: %s' % pprint.pformat(es_response_facets))
     facets = []
-    for name, field in combined_ui_facets.iteritems():
+    for elasticsearch_name, field in combined_ui_facets.iteritems():
```
bfcrampton: nit: `facet_info` over `field`? And same comment about `facet_name` over `elasticsearch_name` to be more general.

api/data_explorer/__main__.py
```diff
@@ -196,10 +195,11 @@ def _process_facets(es):
...
+            'facet'] = elasticsearch_util.get_samples_overview_facet(
```
bfcrampton: nit: Maybe call this key `es_facet` to be a bit more specific?

api/data_explorer/controllers/facets_controller.py
```diff
@@ -79,27 +79,27 @@ def facets_get(filter=None, extraFacets=None):  # noqa: E501
...
-    extra_es_facets, extra_ui_facets = _process_extra_facets(extraFacets)
-    combined_es_facets = OrderedDict(extra_es_facets.items() + current_app.
-                                     config['ELASTICSEARCH_FACETS'].items())
+    extra_ui_facets = _process_extra_facets(extraFacets)
     combined_ui_facets = OrderedDict(extra_ui_facets.items() +
                                      current_app.config['UI_FACETS'].items())
     search = DatasetFacetedSearch(
-        elasticsearch_util.get_facet_value_dict(filter, combined_es_facets),
-        combined_es_facets)
+        elasticsearch_util.get_facet_value_dict(filter, combined_ui_facets),
```
bfcrampton: I would say we should change `filter` since its a built-in function, but this is dictated by the API so I guess it's fine to leave. 

@melissachang?

api/data_explorer/controllers/export_url_controller.py
melissachang: Hmm I prefer `es_field_name`. I'd rather the code apply to the normal, non-Samples Overview case, I think it's easier to read.

api/data_explorer/controllers/facets_controller.py
melissachang: I would leave

api/data_explorer/controllers/export_url_controller.py
bfcrampton: I'd argue that making it easier to read is actually misrepresenting what it is. You can't directly use this as an elastic search field name, because it won't always be one. I'd find it confusing  if that fact were not obvious, and obscuring that makes the code misleading. 

In locations where we are directly extracting `es_field_name` from the facet_info dict, I think it is correct to call it `es_field_name`. Eg:
```
es_field_name = current_app.config['UI_FACETS'][facet_name].get('es_field_name')
```
That field will not be present for the Samples Overview so it is not misrepresenting anything.

melissachang: We shouldn't use `facet_name` because that could be "Facet UI name". So either `es_field_name` or `facet_key`.

So right now we have:
- Facet UI name
- Elasticsearch field name
- Elasticsearch facet object

The problem with `facet_key` is, it's adding another concept that everyone has to understand. And when people write code, they'll have to think "Should this be `facet_key` or `es_field_name`?". And it's going to be impossible to enforce this 100% -- there is going to be `es_field_name` in some places where technically it should be `facet_key`.

Also, every time we use `facet_key` we would have to document "This is es_field_name except for Samples Overview". For example, in api.yaml. While the `facet_key` approach is technically more correct, I think from a pragmatic point of view, it makes the code base harder to understand and maintain.

But if you feel strongly, we can do `facet_key`. :)

api/data_explorer/controllers/facets_controller.py
melissachang: (The [Python style guide](https://github.com/google/styleguide/blob/gh-pages/pyguide.md#3161-names-to-avoid) doesn't say anything about avoiding names that are built-in functions. In general, across languages, I feel like the rule of thumb regarding reserved words is: if it compiles, it's ok.)

api/data_explorer/controllers/facets_controller.py
```diff
@@ -79,27 +79,27 @@ def facets_get(filter=None, extraFacets=None):  # noqa: E501
...
     es_response_facets = es_response.facets.to_dict()
     # Uncomment to print Elasticsearch response python object
-    # current_app.logger.info('Elasticsearch response: %s' % pprint.pformat(es_response_facets))
+    #current_app.logger.info(
+    #    'Elasticsearch response: %s' % pprint.pformat(es_response_facets))
     facets = []
-    for name, field in combined_ui_facets.iteritems():
-        description = field.get('description')
-        es_facet = combined_es_facets[name]
+    for elasticsearch_name, facet_info in combined_ui_facets.iteritems():
```
melissachang: elasticsearch_field_name or es_field_name

api/api.yaml
```diff
@@ -132,6 +132,9 @@ definitions:
       description:
         type: string
         description: Optional facet description.
+      elasticsearch_name:
```
melissachang: elasticsearch_field_name or es_field_name

api/data_explorer/test/mock_responses/api_server_facets_basic.txt
```diff
@@ -1,23 +1,24 @@
...
    "count":1338,
-   "facets":[  
-      {  
+   "facets":[
+      {
          "name":"Region",
          "description":"Region description",
-         "values":[  
-            {  
+         "elasticsearch_name":"Region name",
```
melissachang: Region es field name

Same with Age below

api/data_explorer/controllers/facets_controller.py
```diff
@@ -38,15 +37,16 @@ def _process_extra_facets(extra_facets):
         ui_facet_name = arr[-1]
         field_type = elasticsearch_util.get_field_type(
             es, elasticsearch_field_name)
-        ui_facets[ui_facet_name] = {
-            'elasticsearch_field_name': elasticsearch_field_name,
+        ui_facets[elasticsearch_field_name] = {
+            'ui_facet_name': ui_facet_name,
             'type': field_type
         }
         # TODO(malathir): Figure out how to get description of the field.
```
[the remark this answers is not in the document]
malathi13: No, this is not yet done. This is to show the description on the facet card on UI

api/data_explorer/controllers/facets_controller.py
bfcrampton: I'd disagree, just because something compiles doesn't mean it's the right thing to do. Technically you can define a variable `list = 'foo'` in python, but that means you are overshadowing the built-in `list()` function and it can't be used elsewhere in that context. `filter` is exactly the same, it prevents us from using `filter()` anywhere in this context. Plus it highlights incorrectly in IDEs/Github.

Even though the style guide doesn't explicitly state this, I think where possible we should avoid using built-in function names. In this case it would require changing our API so I'm definitely fine leaving it.

api/data_explorer/controllers/export_url_controller.py
bfcrampton: I think we absolutely **should** document everywhere that it is an es_field_name except for samples overview, regardless of which name we end up choosing. Otherwise it is misleading.

I do still prefer for `facet_key` or `facet_id` as I think it is overall more maintainable to define a new concept rather than overload an existing one and introduce caveats. Say, for example, that we decided to construct the ES facets in the controllers rather than in `__main__` (something we could think about). We cannot simply pass in the key for the facet as the elastic search field, we would need special handling to determine how to construct the particular facet and what the relevant elastic search field(s) are. Additionally, if we allowed people to specify `FiltersFacets` in `ui.json` those would all have the same structure as samples overview and would have to be special cased again. I think right now it might be a little easier to read if it were `es_field_name`, but we should design the API to support these types of things more generally rather than writing in specific code. This is similar to how I initially special cased samples as the only nested facet, but then realized that it's a more general need.

melissachang: SG. How about you do this in a separate PR, since this is a big change, and might (?) involve things outside this PR. I like `facet_id` better than `facet_key`.

bfcrampton: SGTM, I'll updated after this goes in.

PR #188, opened 2018-10-19
Show description of extra facets on UI

malathi13:


api/data_explorer/util/elasticsearch_util.py
```diff
@@ -164,6 +164,25 @@ def get_facet_value_dict(filter_arr, facets):
     return parsed_filter
 
 
+def get_field_description(es, field_name):
+    s = Search(using=es, index=current_app.config['INDEX_NAME'] + '_fields')
+    s.update_from_dict({
```
melissachang: Is it possible to use elasticsearch-dsl? Then it would be easier to maintain, one wouldn't have to deal with lining up brackets.

api/data_explorer/controllers/facets_controller.py
```diff
@@ -37,16 +37,18 @@ def _process_extra_facets(extra_facets):
...
         field_type = elasticsearch_util.get_field_type(
             es, elasticsearch_field_name)
-        ui_facets[elasticsearch_field_name] = {
+        facets[elasticsearch_field_name] = {
             'ui_facet_name': ui_facet_name,
             'type': field_type
         }
-        # TODO(malathir): Figure out how to get description of the field.
-        ui_facets[elasticsearch_field_name][
+        facets[elasticsearch_field_name][
```
melissachang: Delete TODO

malathi13: I think it's deleted.

api/data_explorer/util/elasticsearch_util.py
malathi13: Hmm, Search() is from elasticsearch-dsl. I wasn't able to figure out how to set the query with elasticsearch-dsl.

PR #189, opened 2018-10-19
Truncate the text shown on the chip when an extra facet is selected

malathi13:


malathi13:
screen shot 2018-10-19 at 12 51 51 pm

ui/src/App.js
```diff
@@ -69,12 +69,17 @@ class App extends Component {
             if (field.description) {
               return {
                 label: field.name + " - " + field.description,
-                value: field.elasticsearch_name
+                value: field.elasticsearch_name,
+                chipLabel: (field.name + " - " + field.description).substr(
```
bfcrampton: nit: Could extract `field.name + " - " + field.description` to local var since it's used twice

melissachang:
We should show the entire column name, not just the first 10 characters. See the [SO question Solution 3](https://stackoverflow.com/questions/52482985/react-select-show-different-text-label-for-drop-down-and-control). (Please click "show 5 more comments" to see all comments, that would probably be useful for you.)

PR #190, opened 2018-10-19
Add e2e test for selecting an extra facet

malathi13:


dataset_config/1000_genomes/ui.json
```diff
@@ -69,5 +69,5 @@
         "samples.verily-public-data.human_genome_variants.1000_genomes_sample_info.Main_project_LC_platform"
     }
   ],
-  "enable_field_search": false
+  "enable_field_search": true
```
melissachang: Hmm, good point that this needs to be enabled. In retrospect I didn't put the right order in the spreadsheet.

Let's do these before enabling by default:

- A PR to fix chip text. (First 10 characters looks weird.)
- A PR to fix SQL export, since current SQL is incorrect

Then we can proceed:
- A PR to remove enable_field_search flag. You want to undo relevant parts of #142.
- Then I'll review this PR, and this PR can go in.

ui/tests/e2e/test.js
```diff
@@ -102,6 +110,19 @@ describe("End-to-end", () => {
     await exportToSaturn_selectedCohort();
   });
 
+  test("Field search - chip selected", async () => {
```
melissachang: nit just "Field search"

Typically "- xxx" denotes a variant. In this case, there is no variant, there's just one test.

ui/tests/e2e/test.js
```diff
@@ -102,6 +110,19 @@ describe("End-to-end", () => {
...
+    let initial_select = await page.$x("//div[text()='Select...']");
+    initial_select[0].click();
+    // Click on the 'Avuncular' chip
+    await page.waitForXPath("//div[contains(text(), 'Avuncular')]");
+    let avuncular = await page.$x("//div[contains(text(), 'Avuncular')]");
+    avuncular[0].click();
+    // Assert that the chip was added.
+    let chip = await page.$x("//div[contains(text(), 'Avuncular')]");
+    expect(chip.length).toBe(1);
+  });
```
melissachang: Regarding testing deleting a chip, let's just skip that for now.

ui/tests/e2e/test.js
```diff
@@ -102,6 +110,19 @@ describe("End-to-end", () => {
...
+
```
melissachang: Call assertFacet on Avuncular facet

dataset_config/1000_genomes/ui.json
melissachang: Actually, "PR to remove enable_field_search flag" can come after this PR if you want. I just want the first 2 PRs to be in

PR #191, opened 2018-10-22
Support arbitrary nesting for histogram facets

bfcrampton:


api/data_explorer/util/elasticsearch_util.py
```diff
@@ -15,59 +18,32 @@
...
+    for part in parts:
+        parent = '%s.%s' % (parent, part) if parent else part
+        if parent in current_app.config['NESTED_PATHS']:
+            bucket = bucket.bucket(parent, Nested(path=parent))
+            nestings.append(parent)
+
+    bucket.metric('max', Max(field=field_name))
+    bucket.metric('min', Min(field=field_name))
+
+    print 'SEARCH: %s' % search.to_dict()
```
melissachang: Delete?

As an aside, did you see this output? I thought `print`'s don't show up, only `current_app.logger` do.

bfcrampton: Done.

Hmm I actually did I think, but trying it again it didn't work so maybe not.

PR #192, opened 2018-10-22
Show column name as chip label

malathi13:


ui/src/App.js
```diff
@@ -67,17 +67,16 @@ class App extends Component {
         this.setState({
           fields: data.fields.map(field => {
             if (field.description) {
-              let display_name = field.name + " - " + field.description;
               return {
```
melissachang:
```
# Show different text in chip vs drop-down, since description can be long.
# See https://stackoverflow.com/questions/52482985/react-select-show-different-text-label-for-drop-down-and-control
```

Can you also add this comment to FieldSearch.js, since it's not clear what MultiValue is for? Thanks

PR #196, opened 2018-10-24
Fix SQL query for exporting Samples Overview facet

bfcrampton:
Facets which reference fields from different columns/tables are now handled correctly. Currently this is only the Samples Overview facet, but it will work with any `FiltersFacet` in the future.

Verified the export works for Samples Overview when the fields are in different tables by duplicating the 1000 genomes samples table and referencing one of those fields in the overview facet. The proper `FULL JOIN` statement is generated and the correct number of participants were returned.

### DIFF
--- a/api/data_explorer/controllers/export_url_controller.py
+++ b/api/data_explorer/controllers/export_url_controller.py
@@ -277,11 +277,14 @@ def _get_filter_query(filters):
         for k, v in current_app.config['SAMPLE_FILE_COLUMNS'].iteritems()
     }
 
-    # Key all of the clauses by facet_id and table_name so that they can be AND'd
-    # together and where clauses coalesced by table.
+    # facet_table_clauses must have two levels of nesting (facet_id, table_name)
+    # because clauses from the same column are OR'ed together, whereas clauses
+    # from different columns are AND'ed together.
     facet_table_clauses = {}
     for filter_str in filters:
         splits = filter_str.rsplit('=', 1)
+        # facet_id is "Samples Overview" for the Samples Overview facet, and
+        # es_field_name for all other facets.
         facet_id = splits[0]
         value = splits[1]
         field_type = ''
@@ -299,6 +302,7 @@ def _get_filter_query(filters):
             facet_table_clauses[facet_id][table_name] = []
         facet_table_clauses[facet_id][table_name].append(clause)
 
+    # Map from table name to list of where clauses.
     table_wheres = {}
     table_num = 1
     table_select = '(SELECT %s FROM `%s` WHERE %s)'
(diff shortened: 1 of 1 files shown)


bfcrampton:
Resolves https://github.com/DataBiosphere/data-explorer/issues/197

api/data_explorer/controllers/export_url_controller.py
```diff
@@ -271,61 +271,93 @@ def _get_filter_query(filters):
...
+    table_num = 1
+    table_select = '(SELECT %s FROM `%s` WHERE %s)'
+    query = 'SELECT DISTINCT t1.%s FROM ' % participant_id_column
+
+    def _append_to_query(existing, new, join, table_num):
+        return existing + join if table_num > 1 else existing + new
+
+    # Handle the clauses on a per-facet level.
+    for facet_id in facet_table_clauses.keys():
+        wheres = {}
```
melissachang: table_wheres_for_current_facet

api/data_explorer/controllers/export_url_controller.py
```diff
@@ -271,61 +271,93 @@ def _get_filter_query(filters):
...
-                table_columns[table_name][column] = [clause]
+            facet_id, field_type, value, sample_file_column_fields)
+
+        if facet_id not in facet_table_clauses:
+            facet_table_clauses[facet_id] = {}
+        if table_name not in facet_table_clauses[facet_id]:
+            facet_table_clauses[facet_id][table_name] = []
+        facet_table_clauses[facet_id][table_name].append(clause)
+
+    table_wheres = {}
```
melissachang:
```
# Map from table name to list of where clauses
```

api/data_explorer/controllers/export_url_controller.py
```diff
@@ -271,61 +271,93 @@ def _get_filter_query(filters):
...
+            where = ''
+            for clause in facet_table_clauses[facet_id][table_name]:
+                if len(where) > 0:
+                    where += ' OR (%s)' % clause
+                else:
+                    where = '(%s)' % clause
+            wheres[table_name] = where
+
+        if len(wheres) == 1:
+            # All of the facet where clauses are for the same table, so add
```
melissachang:
```
# This is a normal facet, not "Samples Overview"
```

api/data_explorer/controllers/export_url_controller.py
```diff
@@ -271,61 +271,93 @@ def _get_filter_query(filters):
...
-    for table_name, columns in table_columns.iteritems():
-        table_select = "(SELECT %s FROM `%s` WHERE %s)"
-        where_clause = ""
-        for column, clauses in columns.iteritems():
-            column_clause = ""
-            for clause in clauses:
-                if len(column_clause) > 0:
-                    column_clause = column_clause + " OR "
-                column_clause = column_clause + "(" + clause + ")"
+            # The facet where clauses span multiple tables. In order to OR
```
melissachang:
```
# Samples Overview facet is different. Normally, different columns are AND'ed together.
# Different columns within Samples Overview facet are OR'ed together.
# OR is done using FULL JOIN in case columns are from different tables.
```

api/data_explorer/controllers/export_url_controller.py
```diff
@@ -271,61 +271,97 @@ def _get_filter_query(filters):
...
+                if len(where) > 0:
+                    where += ' OR (%s)' % clause
+                else:
+                    where = '(%s)' % clause
+            table_wheres_current_facet[table_name] = where
+
+        if len(table_wheres_current_facet) == 1:
+            # If all of the facet where caluses are on the same table, add it
+            # to the table_wheres map for coalescing by table below.
+            if table_name not in table_wheres:
```
melissachang: How about adding

```
# All normal, non-Samples Overview facets fall under this case.
```
Whatever you decide is fine, I don't need to re-review.

PR #202, opened 2018-10-31
Don't render a chip when a value is selected

malathi13:


ui/src/components/FieldSearch.js
```diff
@@ -21,8 +12,8 @@ class FieldSearch extends React.Component {
         isMulti="true"
         onChange={this.props.handleChange}
         options={this.props.fields}
-        components={{ MultiValue }}
         value={this.props.extraFacetsOptions}
+        controlShouldRenderValue={false}
```
melissachang:
```
# Don't add chip if we're adding a facet and not selecting any values
```

ui/tests/e2e/test.js
```diff
@@ -239,8 +235,9 @@ describe("End-to-end", () => {
   /**
    * Waits for facet card to be rendered.
    */
-  async function waitForFacetCard(facetName) {
+  async function waitForFacetCard(facetName, facetValue) {
```
melissachang: Why did this change? Just passing "Avuncular" should work.. If it doesn't work, something else needs to be fixed

Issue #205, opened 2018-11-01
Check boxes don't uncheck as expected


RoriCremer:
If you select multiple checkboxes and then try to unselect the first box you selected by clicking on it, it will unselect the last box, not the one you just clicked on.

To reproduce:
In one of the facetCards:
Click the top checkbox so that it is selected
Click the a second checkbox so that it is selected
Click on the top checkbox again (as you would to uncheck it) --note that it is the other checkbox which deselects.

melissachang:
Thanks!

Malathi, can you take a look? https://test-data-explorer.appspot.com/ doesn't have the problem, but master does.

PR #206, opened 2018-11-01
Fix the facet value checking/unchecking bug

malathi13:


melissachang:
Fixes #205 

Malathi, in the future, if a PR fixes an issue, please put "Fixes #xx" in PR description. Then when PR is merged, the issue will automatically be closed.

PR #222, opened 2018-11-15
Format the options in the dropdown

malathi13:


api/api.yaml
```diff
@@ -166,9 +166,12 @@ definitions:
     description: >
       Can be a field (Gender), or a field and selected value (Gender = female).
```
melissachang: Since we're no longer returning "fields", making some more `field` -> `facet` suggestions.

Can represent a facet (Gender), or a facet and selected value (Gender = female).

api/api.yaml
```diff
@@ -166,9 +166,12 @@ definitions:
...
     properties:
-      display_text:
+      facet_name:
         type: string
-        description: The text displayed in the dropdown.
+        description: The name of the facet.
+      facet_description:
+        type: string
+        description: The description of the facet.
```
melissachang: Optional. The description of the facet.

melissachang: Lines 181-182:

           If this search result represents a facet, facet_value is the empty string.
           If this search result represents a facet and selected value, facet_value is the selected value.

ui/src/App.js
```diff
@@ -70,8 +70,9 @@ class App extends Component {
         this.setState({
           searchResults: data.search_results.map(searchResult => {
             return {
-              label: searchResult.display_text,
-              value: searchResult.elasticsearch_field_name,
+              facetName: searchResult.facet_name,
```
melissachang: Update the [comments here](https://github.com/DataBiosphere/data-explorer/blob/776cc4667d5df287a6ba8537e7d8f775f2126903/ui/src/App.js#L36).
Doesn't need to be long, just something like:
```
      // Search results shown in the search drop-down.
      // This is an array of dicts, where each dict has facetName, facetDescription, esFieldName, and facetValue
```
By convention, let's document unobvious state variables in the state initialization section.

ui/src/components/Search.js
```diff
@@ -5,13 +5,42 @@ class Search extends React.Component {
   constructor(props) {
     super(props);
   }
+  renderOption = option => {
+    if (option.facetDescription != null) {
+      return (
+        <div>
+          <span style={{ color: "grey" }}>Add</span>
```
melissachang: grey -> silver

Throughout whole file

ui/src/components/Search.js
```diff
@@ -5,13 +5,42 @@ class Search extends React.Component {
   constructor(props) {
     super(props);
   }
+  renderOption = option => {
```
melissachang:
```
// renderOption is used to render the drop-down
```

ui/src/components/Search.js
```diff
@@ -5,13 +5,42 @@ class Search extends React.Component {
...
+      return (
+        <div>
+          <span style={{ color: "grey" }}>Add</span>
+          <span style={{ color: "black" }}> {option.facetName} </span>
+          <span style={{ color: "grey" }}>facet</span>
+        </div>
+      );
+    }
+  };
+  renderValue = option => {
```
melissachang:
```
// renderValue is used for autocomplete. If I type "foo" into search box,
// drop-down options whose renderValue contains "foo" will be shown in the drop-down.
```

ui/src/App.js
malathi13: Done

ui/src/components/Search.js
malathi13: Done

ui/src/components/Search.js
malathi13: Done

ui/tests/e2e/test.js
```diff
@@ -115,9 +115,11 @@ describe("End-to-end", () => {
     // Click on the drop down
     let initial_select = await page.$x("//div[text()='Select...']");
     initial_select[0].click();
-    // Click on the 'Avuncular' chip
-    await page.waitForXPath("//div[contains(text(), 'Avuncular')]");
-    let avuncular = await page.$x("//div[contains(text(), 'Avuncular')]");
+    // Click on the 'Avuncular' option
+    await page.waitForXPath("//span[contains(normalize-space(), 'Avuncular')]");
+    let avuncular = await page.$x(
+      "//span[contains(normalize-space(), 'Avuncular')]"
```
melissachang: I looked into why `"//span[contains(normalize-space(), 'Avuncular')]"` works and `$x("//span[contains(text(), 'Avuncular')]")` doesn't. This [SO post](https://stackoverflow.com/questions/3655549/xpath-containstext-some-string-doesnt-work-when-used-with-node-with-more) explains why.

For future reference, this should be `//span[text()[contains(.,'Avuncular')]]`.

PR #228, opened 2018-11-19
Fail script if command fails

### LINKED ISSUE #224: deploy/deploy-api.sh: line 20: jq: command not found
I was deploying my data explorer:

```
$ deploy/deploy-api.sh my_config
deploy/deploy-api.sh: line 20: jq: command not found
Deploying my_config API Server to project 

ERROR: (gcloud.config.set) argument VALUE: Must be specified.
Usage: gcloud config set SECTION/PROPERTY VALUE [optional flags]
  optional flags may be  --help | --installation
```

but it continued running. Fortunately, my cloud project was already set properly.

The `deploy-api.sh` script should include at the top:

```
set -o errexit
set -o nounset
```

This would cause the script to exit immediately:

```
$ deploy/deploy-api.sh my_config
deploy/deploy-api.sh: line 23: jq: command not found
$
```

melissachang:
Fixes #224


Dockerfile
```diff
@@ -0,0 +1,11 @@
+# This must be run in project root, to pick up dataset_config/.
```
bfcrampton: Isn't this file in api/Dockerfile? Why do we need to duplicate?

app.yaml
```diff
@@ -0,0 +1,9 @@
+runtime: custom
```
bfcrampton: Same, a similar file in ui/app.yaml (with template)?

Dockerfile
melissachang: Oops, thanks for catching this. deploy/deploy-api.sh moves these files here temporarily. I ran deploy-api.sh for testing this PR, canceled it, and forgot to delete these files. (We should probably think of something less hacky.)

Deleted Dockerfile and app.yaml from this PR.

PR #236, opened 2018-11-28
Add dataset config to set search placeholder text

melissachang:


ui/src/App.js
```diff
@@ -140,7 +144,8 @@ class App extends Component {
         console.error(error);
       } else {
         this.setState({
-          datasetName: data.name
+          datasetName: data.name,
+          searchPlaceholderText: data.search_placeholder_text
```
malathi13: I think this will throw an error if the config does not have "search_placeholder_text". Maybe set a default text in the api.yaml, or set a default when returning the response from dataset controller.

melissachang: I tried deleting search_placeholder_text from ui.json. The search box shows "Select...". I assume data.search_placeholder_text is empty string by default.

I think react-select is being smart about this -- if placeholder is empty string, it will show "Select..." instead of empty string.

Btw, I restored test_dataset_controller.py. It was a bit useful... I had to change the code to make sure things worked without search_placeholder_text in ui.json.

PR #239, opened 2018-11-29
Remove align=start from FacetCard

melissachang:
With align=start:
with_align_start

Without align=start:
without_align_start

I added `align=start` to make the checkbox look better when the facet value text wraps, such as with `Utah residents`.

However, for one-line facet lines, `align=start` causes the facet value text to be a pixel or two higher than it should be. I'd rather have the one-line case look right.


bfcrampton:
Personally I like it centered on the multiline facet values better anyways 😄

PR #240, opened 2018-12-03
Fix bucketing for facets with 0 and 1

melissachang:
Before:
before

After:
after

### DIFF
--- a/api/data_explorer/util/elasticsearch_util.py
+++ b/api/data_explorer/util/elasticsearch_util.py
@@ -48,7 +48,7 @@ def _get_field_range(es, field_name):
 def _get_bucket_interval(es, field_name, field_type):
     field_range = _get_field_range(es, field_name)
     # If an integer field has 0s and 1s, return 1 instead of .1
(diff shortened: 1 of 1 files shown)


api/data_explorer/util/elasticsearch_util.py
```diff
@@ -45,9 +45,11 @@ def _get_field_range(es, field_name):
     return (aggs['max']['value'] - aggs['min']['value'])
 
 
-def _get_bucket_interval(es, field_name):
+def _get_bucket_interval(es, field_name, field_type):
     field_range = _get_field_range(es, field_name)
-    if field_range <= 1:
+    # If an integer field has 0s and 1s, return 1 instead of .1
+    is_integer_field = field_type == 'long' or field_type == 'integer' or field_type == 'short' or field_type == 'byte'
```
bfcrampton: nit: could shorten to:
```
is_integer_field = field_type in ['long', 'integer', 'short', byte']
```

melissachang: Done

PR #247, opened 2018-12-11
Add flag to turn on the free text search feature

malathi13:


dataset_config/template/ui.json
```diff
@@ -6,6 +6,10 @@
 // facets:
 //   Facets to be shown in the UI. Facets will appear in the order they are
 //   listed here. ui_facet_description is optional.
+// enable_search:
```
melissachang: enable_search_values:  or enable_search_facet_values:
  Enable searching on facet values.

dataset_config/1000_genomes/ui.json
```diff
@@ -8,6 +8,7 @@
 //   listed here. ui_facet_description is optional.
 {
   "search_placeholder_text": "Try typing exome or center",
+  "enable_search": false,
```
melissachang: Let's delete. We want to make sure things work if ui.json doesn't have this flag, so we can test that out with 1000 Genomes.

api/api.yaml
```diff
@@ -121,6 +121,8 @@ definitions:
       search_placeholder_text:
         type: string
         description: Optional. What to show in the search box by default
+      enable_search:
```
melissachang: enable_search_values or enable_search_facet_values
(Right now we allow search on facet name and description)

api/data_explorer/test/test_dataset_controller.py
```diff
@@ -15,7 +15,7 @@ def create_app(self):
 
     def test_dataset_get(self):
         """Test case for dataset_get"""
-        expected_response = {'name': dataset_name}
+        expected_response = {'name': dataset_name, 'enable_search': False}
```
melissachang: Let's revert this change, we want to make sure things work if ui.json doesn't have this flag

melissachang: Got it

PR #250, opened 2018-12-13
Upgrade Elasticsearch

melissachang:
Makes debugging https://github.com/elastic/elasticsearch/issues/36452 a little easier.

Also update search_placeholder_text to a more user-friendly string.


melissachang:
Upgrading ES in production is very difficult, since we're using pires

PR #251, opened 2018-12-14
e2e test for search on values

malathi13:


ui/tests/e2e/test.js
```diff
@@ -168,6 +168,27 @@ describe("End-to-end", () => {
     await waitForFacetsUpdate(3500);
   });
 
+  test("Search box - values", async () => {
```
melissachang: Search box - select row with facet value

melissachang: Delete `test("Search box - click on drop-down row")`. This test already clicks on drop-down row.

ui/tests/e2e/test.js
```diff
@@ -168,6 +168,27 @@ describe("End-to-end", () => {
...
+    // Select first result
+    await page.waitForXPath("//span[contains(text(), 'Add')]");
+    let results = await page.$x("//span[contains(text(), 'Add')]");
+    await results[0].click();
+
+    // Assert Relationship facet is added
+    await waitForFacetCard("Relationship");
+    await assertFacet("Relationship", "1", "mother", "831");
+
+    // Make sure non-selected facet values are gray.
```
melissachang: opt - These tests should really be asserting that a facet value is checked, not that a different value is gray. I could have sworn that was somewhere in this test, but I can't find it now. Up to you -- add assertFacetValueSelect() and have this test and others use it.

malathi13: I am not sure which other tests need this, I changed this test to check if the value is selected.

PR #256, opened 2019-01-03
After faceted search, rerender in one stage instead of two

### LINKED ISSUE #63: After faceted search, rerender in one stage instead of two
To best illustrate this issue, put `import time; time.sleep(1)` into facets_controller.py facets_get().

After clicking on a checkbox, rerendering happens in two stages:

1. Checkbox has checkmark, unchecked Age values turn gray
stage1

2. New facet data returned from backend. All numbers outside of Age facet update. (Age total count also updates.)
stage2

Two stages is jarring. Rerendering should happen in one stage. We should hold off on 1) until after backend RPC completes.

AFAICT, [Boardwalk](https://ucsc-cgp.org/boardwalk) doesn't have this problem.

wnojopra:
Fixes #63 .
Before:
https://screencast.googleplex.com/cast/NTE1MTQ1NjEyMjM3MjA5NnwwZTlhYjk1NS0zZg
After:
https://screencast.googleplex.com/cast/NTA3MDY0NjA0NDI2MjQwMHxhMzk4MzZmMC0zMw
(The previous two screencasts include a 1 second sleep in the api to represent a slower query).

Previously, after doing a faceted search, the app would do two state changes. The first state change happens immediately after clicking, updating the selectedFacetValues. The second state change happens once we get a response from the api, updating the entire facetsMap. If there is a longish delay while waiting for the response, this results in a somewhat jarring ui effect. 

The idea to fixing this:
1) Introduce a new state in App.js called 'awaitingApi'. This is a boolean meant to represent whether or not the app is waiting for a response from the API. 
2) If awaitingApi is true, do not re-render. This is checked in shouldComponentUpdate, which is invoked before rendering when new states are being received.
3) Update FacetCard.js to update the summed facet values on re-render instead of on prop change. This is necessary because previously, the two render steps allowed for a prop change check to happen in between. I also like changing this because componentWillReceiveProps was deprecated earlier this year.

### DIFF
--- a/ui/src/App.js
+++ b/ui/src/App.js
@@ -60,7 +60,8 @@ class App extends Component {
       // facetDescription - The description of the facet.
       // esFieldName - The elasticsearch field name of the facet.
       // facetValue
-      searchResults: []
+      searchResults: [],
+      awaitingApi: false
     };
 
     this.apiClient = new ApiClient();
@@ -68,12 +69,14 @@ class App extends Component {
     this.facetsApi = new FacetsApi(this.apiClient);
     this.facetsCallback = function(error, data) {
       if (error) {
+        this.setState({ awaitingApi: false });
         console.error(error);
         // TODO(alanhwang): Redirect to an error page
       } else {
         this.setState({
           facets: this.getFacetMap(data.facets),
-          totalCount: data.count
+          totalCount: data.count,
+          awaitingApi: false
         });
       }
     }.bind(this);
@@ -121,6 +124,10 @@ class App extends Component {
     this.updateFacets = this.updateFacets.bind(this);
     this.handleSearchBoxChange = this.handleSearchBoxChange.bind(this);
     this.loadOptions = this.loadOptions.bind(this);
+  }
+
+  shouldComponentUpdate(nextProps, nextState) {
+    return !nextState.awaitingApi;
   }
 
   render() {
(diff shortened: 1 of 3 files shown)


ui/src/App.js
```diff
@@ -60,20 +60,23 @@ class App extends Component {
       // facetDescription - The description of the facet.
       // esFieldName - The elasticsearch field name of the facet.
       // facetValue
-      searchResults: []
+      searchResults: [],
+      awaitingApi: false
```
melissachang: awaitingFacetsApi

ui/src/components/facets/FacetCard.js
```diff
@@ -123,7 +111,10 @@ class FacetCard extends Component {
         <Typography>{this.props.facet.name}</Typography>
         {this.props.facet.name != "Samples Overview" ? (
           <Typography className={classes.totalFacetValueCount}>
-            {this.totalFacetValueCount}
+            {this.sumFacetValueCounts(
```
melissachang: Thanks for this change. This is better react style anyways.

wnojopra:
@melissachang I have deployed these UI changes to test-data-explorer and nhs-explorer. Please take a look. I will not merge until I hear back from you.

melissachang:
They look great. Thanks, this is **so** much better.

PR #263, opened 2019-01-30
Fix intervals for numeric facets

melissachang:
Current intervals are confusing, it's unclear if 2 is in the 0-2 bucket or the 2-4 bucket:

interval

**Solution 1**

0-1
2-3
4-5

The downside is that technically 1.5 doesn't belong to any bucket, but practically speaking, I think it's obvious where it goes.

**Solution 2 - Interval notation**

[0,2)
[2,4)
[4,6)

This is mathematically precise, but it's ugly, requires some thinking/is not as easy to parse, and not everyone knows interval notation off the top of their heads. So I'm doing Solution 1.

---

Solution 1 doesn't address this case:

.6-.7
.7-.8
.8-.9

But I still think Solution 1 is preferable to Solution 2 overall.

CC @deflaux


melissachang:
Added test. Btw, can you review #262? I need to merge that first

PR #245, opened 2018-12-10
Initial implementation of histogram facet visualizations

bfcrampton:
First stab at https://github.com/DataBiosphere/data-explorer/issues/9

image

PR #267, opened 2019-02-04
Add HistogramFacet for visualizations

bfcrampton:
This is a reopening of https://github.com/DataBiosphere/data-explorer/pull/245 since it was cleaner to start a new branch instead of rebasing. All of the comments in that PR should be addressed here, including switching to Vega Lite which is much nicer. Also added highlighting off selected facets, and made categorical facets horizontal.

screen shot 2019-02-04 at 12 18 55 pm


ui/src/components/facets/HistogramFacet.js
```diff
@@ -0,0 +1,113 @@
...
+        field: "count",
+        type: "quantitative",
+      }
+    ],
+    x: {},
+    y: {}
+  }
+}
+
+function isRange(name) {
```
bfcrampton: We may want to just add an `es_field_type` field to the `Facet` API object instead of doing this regex matching.

ui/package.json
```diff
@@ -5,13 +5,19 @@
...
     "@material-ui/icons": "^2.0.3",
+    "@material-ui/lab": "^3.0.0-alpha.29",
     "axios": "^0.18.0",
     "data_explorer_service": "file:src/api",
     "lodash.debounce": "^4.0.8",
     "react": "^16.3.0",
     "react-dom": "^16.3.0",
     "react-scripts": "^1.1.4",
-    "react-select": "^2.0.0"
+    "react-select": "^2.0.0",
```
melissachang: Alphabetize

ui/package.json
```diff
@@ -5,13 +5,19 @@
...
+    "vega-lib": "4.4.0",
```
melissachang: Let's add ^, like the rest of this file

ui/src/components/facets/FacetCard.js
```diff
@@ -1,182 +1,30 @@
...
-      facetValue,
-      isSelected
-    );
-  }
+import TextFacet from "components/facets/TextFacet";
+import HistogramFacet from "components/facets/HistogramFacet";
+
+function FacetCard(props) {
+  const { classes } = props;
+  const rand = Math.random() < 0.5;
```
melissachang: Delete.

Actually, this class doesn't do much. Delete this file and inline in FacetsGride.js.

ui/src/components/facets/HistogramFacet.js
```diff
@@ -0,0 +1,113 @@
...
+  encoding: {
+    color: {
+      field: "dimmed",
+      type: "nominal",
+      scale: {
+        range: ["#4c78a8", "#aaafb7"]
+      },
+      legend: null,
+    },
+    tooltip: [
```
melissachang: Replace 21-32:
```
    tooltip: [
      {
        title: "facet_value",
        field: "count",
        type: "quantitative",
      },
    ],
```
I want tooltip to be "female: 1760"

ui/src/components/facets/HistogramFacet.js
```diff
@@ -0,0 +1,113 @@
...
+
+    const spec = Object.assign({}, baseSpec);
+    // Range facets are shown as vertical histograms, while categorical
+    // ones are horizontal in order to allow for more space for text.
+    if (rangeFacet) {
+      spec.encoding.x = {
+        field: "facet_value",
+        type: "nominal",
+        title: this.props.facet.name,
+        sort: sortOrder,
```
melissachang: It's more idiomatic javascript to inline. Delete 54, 62. This line: `sort: this.facetValues.map(v => v.name)`

Same with `data`. 51-64 should be deleted.

ui/src/components/facets/HistogramFacet.js
melissachang: I agree, can you do that in this PR?

ui/src/components/facets/TextFacet.js
```diff
@@ -0,0 +1,182 @@
...
+          {this.props.facet.description}
+        </Typography>
+        <List dense className={classes.facetValueList}>
+          {facetValueDivs}
+        </List>
+      </div>
+    );
+  }
+
+  isDimmed(facetValue) {
```
melissachang: Let's call this isValueDimmed. Same with HistogramFacet.js.

ui/src/components/facets/HistogramFacet.js

I was only able to do this by adding a 'text' field to the data with this format. Unfortunately the [tooltip spec](https://vega.github.io/vega-lite/docs/encoding.html#text) doesn't allow you use two fields in the title like that.

bfcrampton:
Addressing some of your comments:
> * For numeric facets, it would be nice if x-axis legend text is horizontal instead of vertical. Could the labels be scattered? Or perhaps we should be smarter about the labels; instead of 0-9/10-19/20-29, we could somehow display 0/10/20.

Did this in this PR. Also added the `labelsOverlap = true` as described [here](https://vega.github.io/vega-lite/docs/axis.html#labels) which looks pretty nice I think. For the Total Exome Sequence ones, without it the labels overlap a lot and are not visible.
> * Title should look the same for TextFacet and HistogramFacet: Not bold, top (for both string and numeric facets), same font size, etc
> * HistogramFacet should look like a card, similar to TextFacet. Use `card` style and background should be white.
> * For facets with many values, like Population, instead of shrinking the bar, we should add a scrollbar, similar to TextFacet

This shouldn't be too hard, just apply the card wrapper (with title) around the histogram and make it scrollable I think? 
> * I don't see a title for Total Exome Sequence facet.

Hmm I haven't seen this. maybe the horizontal labels fixed this?
> * For long facet values (like Population facet), can text be wrapped so entire string can be shown?

Definitely something we should do, not sure if vega supports that or not yet.
> * Bars for Population facet are all the same width. There is some cutoff problem. You also see it with Samples Overview facet, first 5 values.

I see this as well (if you expand your window large or zoom out on the page you can actually see this fixed). This is because of the fixed I applied (it looks really bad without this). Vega renders the equivalent of a static image so it doesn't scale with the page at all. There is an [`autosize`](https://vega.github.io/vega-lite/docs/size.html#autosize) property I experimented with a little bit, but couldn't get it looking good. We may want to revisit that.
> * On a related note, there should be a max width for the bar. Facets with just one value should not have a humongous bar.

I don't quite know what you mean by this, just that there should be space around the bar to be clear what it is?
> * For Total Exome Sequence facet, it's impossible to get tooltip to show for 80-90B. Can vega show tooltip if I'm anywhere in that column?

I think this is doable if we also put invisible data in to trigger the tooltip anywhere on the screen (I believe we discussed this a while ago). I haven't tried that yet, but doesn't seem too hard.

api/api.yaml
```diff
@@ -151,6 +151,9 @@ definitions:
       es_field_name:
         type: string
         description: The Elasticsearch field name.
+      es_field_type:
+        type: string
+        description: The type of Elasticsearch field.
```
melissachang: nit

The Elasticsearch field type.

api/data_explorer/controllers/facets_controller.py
```diff
@@ -69,7 +67,8 @@ def facets_get(filter=None, extraFacets=None):  # noqa: E501
     :type extraFacets: List[str]
     :rtype: FacetsResponse
     """
-    _process_extra_facets(extraFacets)
+    es = Elasticsearch(current_app.config['ELASTICSEARCH_URL'])
```
melissachang: Why did this change? I don't see `es` being used anywhere.

api/data_explorer/controllers/facets_controller.py
```diff
@@ -86,9 +85,12 @@ def facets_get(filter=None, extraFacets=None):  # noqa: E501
     #    'Elasticsearch response: %s' % pprint.pformat(es_response_facets))
     facets = []
     for es_field_name, facet_info in combined_facets.iteritems():
+        current_app.logger.info(
```
melissachang: Delete

api/data_explorer/controllers/facets_controller.py
```diff
@@ -86,9 +85,12 @@ def facets_get(filter=None, extraFacets=None):  # noqa: E501
...
+            'FIELD: %s INFO: %s' % (es_field_name, facet_info))
         name = facet_info.get('ui_facet_name')
         description = facet_info.get('description')
         es_facet = facet_info.get('es_facet')
+        field_type = facet_info.get('type', 'text')
```
melissachang: Remove `text`. It's a bug if facet_info doesn't have `type`.

api/data_explorer/controllers/facets_controller.py
```diff
@@ -86,9 +85,12 @@ def facets_get(filter=None, extraFacets=None):  # noqa: E501
     #    'Elasticsearch response: %s' % pprint.pformat(es_response_facets))
     facets = []
     for es_field_name, facet_info in combined_facets.iteritems():
+        current_app.logger.info(
+            'FIELD: %s INFO: %s' % (es_field_name, facet_info))
         name = facet_info.get('ui_facet_name')
         description = facet_info.get('description')
```
melissachang: Inline 90, 91, 93

dataset_config/1000_genomes/ui.json
```diff
@@ -89,5 +89,5 @@
     }
   ],
   "enable_search_values": true,
-  "show_viz_toggle": false
+  "show_viz_toggle": true
```
melissachang: Revert.

(For local runs, I create a separate 1000_genomes_viz directory)

ui/src/components/facets/HistogramFacet.js
```diff
@@ -0,0 +1,99 @@
...
+class FacetHistogram extends Component {
+  constructor(props) {
+    super(props);
+    this.facetValues = this.props.facet.values;
+    this.isValueDimmed = this.isValueDimmed.bind(this);
+  }
+
+  render() {
+    const spec = Object.assign({}, baseSpec);
+    // Categorical  facets are shown as horizontal histograms,
```
melissachang: nit change double space to single space

api/api.yaml
bfcrampton: Done.

api/data_explorer/controllers/facets_controller.py
bfcrampton: Sorry was going to use elasticsearch.get_field_type here but the facet_info type ended up being sufficient. Reverted that part.

api/data_explorer/controllers/facets_controller.py
bfcrampton: Done.

api/data_explorer/controllers/facets_controller.py
bfcrampton: Done.

api/data_explorer/controllers/facets_controller.py
bfcrampton: Done.

dataset_config/1000_genomes/ui.json
bfcrampton: Done.

ui/src/components/facets/HistogramFacet.js

PR #270, opened 2019-02-06
Support selection of histogram type facets

bfcrampton:


ui/src/components/facets/HistogramFacet.js
```diff
@@ -2,8 +2,6 @@ import React, { Component } from "react";
 import VegaLite from "react-vega-lite";
```
melissachang: Looks like we're only using react-vega-lite and vega-tooltip. Can you delete everything from package.json (react-vega, vega-embed, vega-lib, vega-lite)?

ui/src/components/facets/HistogramFacet.js
```diff
@@ -88,6 +95,35 @@ class FacetHistogram extends Component {
       !this.props.selectedValues.includes(facetValue.name)
     );
   }
+
+  onClick(event, item) {
+    // Ignore clicks which are not located on histogram
+    // bars.
+    if (item != null) {
```
melissachang:
```
if (item && item.datum) {
```
If I click inside the graph area but not on a bar, item is set and item.datum is not set.

ui/src/components/facets/HistogramFacet.js
```diff
@@ -88,6 +95,35 @@ class FacetHistogram extends Component {
...
+      // facetValue is a string, eg "female"
+      const facetValue = item.datum.facet_value;
+      let isSelected;
+      if (
+        this.props.selectedValues != null &&
```
melissachang: // this.props.selectedValues contains what was selected before the click.
// isSelected contains if facet value was selected after the click.

ui/src/components/facets/FacetsGrid.js
```diff
@@ -17,6 +17,7 @@ function facetCard(props, facet) {
     return (
       <HistogramFacet
         facet={facet}
+        updateFacets={props.updateFacets}
```
melissachang: This has the same problem as the tooltip -- it's not possible to select values with low numbers. Maybe if we do the transparent background bar, that could help with this.

ui/src/components/facets/HistogramFacet.js

ui/src/components/facets/HistogramFacet.js

ui/src/components/facets/HistogramFacet.js

ui/src/components/facets/FacetsGrid.js
bfcrampton: Noted, I'll fix both in my next PR.

bfcrampton:
It turns out that since react-vega-lite has a [peerDependency](https://nodejs.org/en/blog/npm/peer-dependencies/#using-peer-dependencies) [on vega-lite](https://github.com/vega/react-vega-lite/blob/master/package.json#L16), and react-vega [on vega-lib](https://github.com/vega/react-vega/blob/master/package.json#L16), they also have to be present on our `package.json`. react-vega-lite has a direct dependency on react-vega though so kept that out.

ui/package.json
```diff
@@ -13,11 +13,10 @@
     "react-dom": "^16.3.0",
     "react-scripts": "^2.0.0",
     "react-select": "^2.0.0",
-    "react-vega": "^4.0.2",
     "react-vega-lite": "^2.0.2",
-    "vega-embed": "^3.29.0",
     "vega-lib": "^4.4.0",
-    "vega-lite": "^3.0.0-rc12"
+    "vega-lite": "^3.0.0-rc12",
```
melissachang: Add a comment for why these two are here, so someone else in the future doesn't make the same mistake. See https://stackoverflow.com/a/14221781/6447189

ui/package.json
```diff
@@ -46,5 +45,11 @@
   },
   "jest": {
     "preset": "jest-puppeteer"
-  }
+  },
+  "browserslist": [
```
melissachang: Why did this change?

ui/package.json
bfcrampton: Done.

ui/package.json
bfcrampton: Huh, not sure. Reverted.

ui/package.json
```diff
@@ -29,6 +28,7 @@
     "eject": "react-scripts eject"
   },
   "//": [
+    "vega-lib and vega-lite are peerDependencies of react-vega-lite.",
```
melissachang: Is it possible to split this into two comments? Leave 32 here, move 31 to 17.

bfcrampton: Yup it is, done.

PR #272, opened 2019-02-11
Expand HistogramFacet tooltip/selection area 

bfcrampton:
Precursor PR: https://github.com/DataBiosphere/data-explorer/pull/270

I spent a while investigating alternatives for implementing this functionality. From what I understand about Vega selections, there doesn't seem to be an easy way to expand the clickable area for a facet bar, short of doing manual click location calculations. It's even more difficult for the tooltip since we don't have any control over what event(s) triggers the tooltip, we can only customize it's look and provide the actual `Handler` for customizing the UI.

This solution ended up being quite simple to implement. That being said, it is still a bit 'hacky' so I can post an issue about this on the vega repo if we'd still like to look for other options.

### DIFF
--- a/ui/src/components/facets/HistogramFacet.js
+++ b/ui/src/components/facets/HistogramFacet.js
@@ -5,7 +5,7 @@ import { Handler } from "vega-tooltip";
 const baseSpec = {
   $schema: "https://vega.github.io/schema/vega-lite/v3.json",
   mark: "bar",
-  width: 400,
+  width: 250,
   height: 300,
   encoding: {
     color: {
@@ -21,14 +21,23 @@ const baseSpec = {
       type: "nominal"
     },
     x: {},
-    y: {}
+    y: {},
+    opacity: {
+      field: "opaque",
+      type: "nominal",
+      scale: {
+        range: [0, 1]
+      },
+      legend: null
+    }
   }
 };
 
-const count_axis = {
+const countAxis = {
   field: "count",
   type: "quantitative",
-  title: ""
+  title: "",
+  stack: null
 };
 
 function isCategorical(facet) {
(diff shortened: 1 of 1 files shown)


bfcrampton:
Oops sorry, added to the other

ui/src/components/facets/HistogramFacet.js
```diff
@@ -73,11 +82,29 @@ class FacetHistogram extends Component {
...
           count: v.count,
           dimmed: this.isValueDimmed(v),
-          text: `${v.name}: ${v.count}`
+          text: `${v.name}: ${v.count}`,
+          opaque: true
         };
       })
     };
 
+    // Add an additional copy of the with a count equal to the
```
melissachang: Replace 91-94

```
// Create transparent bar that extends the entire length of the cart. This makes tooltip/selection easier for
// facet values that have very low count.
```

ui/src/components/facets/HistogramFacet.js
```diff
@@ -21,14 +21,23 @@ const baseSpec = {
       type: "nominal"
     },
     x: {},
-    y: {}
+    y: {},
+    opacity: {
```
melissachang:
```
# opacity is needed for creating transparent bars
```

ui/src/components/facets/HistogramFacet.js
```diff
@@ -73,11 +82,29 @@ class FacetHistogram extends Component {
...
+          opaque: true
         };
       })
     };
 
+    // Add an additional copy of the with a count equal to the
+    // maximum count in this facet. This data is transparent, but
+    // makes it easier to select/hover tooltip for facets with
+    // very small bars.
+    const maxCount = data.values.reduce(
```
melissachang:
```
const maxFacetValue = Math.max(...data.values.map(v => v.count))
```

ui/src/components/facets/HistogramFacet.js

ui/src/components/facets/HistogramFacet.js

ui/src/components/facets/HistogramFacet.js

PR #273, opened 2019-02-12
Some HistogramFacet styling

melissachang:
- Render card
- Make title same format as TextFacet (top center)
- Lato font


melissachang:
Just realized this broke facet selection. (I'll fix it.) Eventually we should have an e2e test. But let's hold off until the flag is on by default.

ui/src/components/facets/HistogramFacet.js
```diff
@@ -47,7 +65,7 @@ function isCategorical(facet) {
   );
 }
 
-class FacetHistogram extends Component {
+class HistogramFacet extends Component {
```
bfcrampton: Oops :)

PR #274, opened 2019-02-12
More histogram facet styling

melissachang:
- Update bar and header to use Terra colors from Lou's screens
- Fixed bar width
- Create FacetHeader component and use it in both TextFacet and HistogramFacet

### DIFF
--- a/ui/src/components/facets/FacetHeader.js
+++ b/ui/src/components/facets/FacetHeader.js
@@ -52,7 +52,7 @@ class FacetHeader extends Component {
    * @param facetValues FacetValue[] to sum counts over
    * @param selectedValueNames Optional string[] to select a subset of facetValues to sum counts for
    * @return number count the total sum of all facetValue counts, optionally filtered by selectedValueNames
-   * */
+   */
   sumFacetValueCounts(facetValues, selectedValueNames) {
     let count = 0;
     if (selectedValueNames == null || selectedValueNames.length === 0) {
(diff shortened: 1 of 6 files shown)


ui/src/components/facets/FacetHeader.js
```diff
@@ -0,0 +1,73 @@
...
+        </Typography>
+      </React.Fragment>
+    );
+  }
+
+  /**
+   * @param facetValues FacetValue[] to sum counts over
+   * @param selectedValueNames Optional string[] to select a subset of facetValues to sum counts for
+   * @return number count the total sum of all facetValue counts, optionally filtered by selectedValueNames
+   * */
```
bfcrampton: micro nit: remove extra `*`?

ui/src/components/facets/FacetHeader.js
```diff
@@ -0,0 +1,73 @@
...
+  sumFacetValueCounts(facetValues, selectedValueNames) {
+    let count = 0;
+    if (selectedValueNames == null || selectedValueNames.length === 0) {
+      facetValues.forEach(value => {
```
bfcrampton: nit: do you need the brackets?

```
facetValues.forEach(value => count += value.count);
```

ui/src/components/facets/FacetHeader.js
melissachang: Done

ui/src/components/facets/FacetHeader.js
melissachang: Maybe not, but I think this is cleaner -- it matches 64-66

PR #276, opened 2019-02-13
Add snackbar with HistogramFacet tips

melissachang:
localStorage is used to only show the snackbar once.


melissachang:
Just the two tips. Are there others you think should be included?

PR #277, opened 2019-02-13
Minor UI changes from Sean

melissachang:
- Make all bars horizontal. Then one doesn't have to wonder why some are horizontal and some are vertical.
- Make search box initial text more verbose, to be more clear.

### DIFF
--- a/ui/src/components/Header.js
+++ b/ui/src/components/Header.js
@@ -28,8 +28,9 @@ const styles = {
   },
   toggleButtonGroup: {
     borderRadius: "20px",
-    // Must be same color as appBar background to fix border-radius bleed
-    // (https://stackoverflow.com/a/30356787/6447189)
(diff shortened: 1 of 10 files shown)


ui/src/components/Header.js
```diff
@@ -23,7 +28,7 @@ const styles = {
   },
   toggleButtonGroup: {
     borderRadius: "20px",
-    // Must be same color as header background to fix border-radius bleed
+    // Must be same color as appBar background to fix border-radius bleed
     // (https://stackoverflow.com/a/30356787/6447189)
     backgroundColor: "#55a146"
```
wnojopra: I think in general we should have constants somewhere for values like this. Would be easy to keep track of changes too.

melissachang: Unfortunately the appBar background is a gradient. I used the midpoint color here. Added a comment.

PR #281, opened 2019-02-19
Fix bucketing for facets with 0.0, 1.0

melissachang:
Before:
before

After:
after

#240 fixed facets with integers (0, 1), but not floats (0.0, 1.0). So try a different approach that works with both floats and integers.

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -48,5 +48,11 @@
   },
   "jest": {
     "preset": "jest-puppeteer"
-  }
+  },
+  "browserslist": [
+    ">0.2%",
+    "not dead",
+    "not ie <= 11",
+    "not op_mini all"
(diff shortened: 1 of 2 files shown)


melissachang:
The package.json happened automatically, let's check that in also.

ui/package.json
```diff
@@ -48,5 +48,11 @@
   },
   "jest": {
     "preset": "jest-puppeteer"
-  }
+  },
```
wnojopra: I noticed that changes from this file tend to show up, but I always end up not committing them. What is this browserslist change? Should we commit it? Should we add it to the .gitignore?

wnojopra: edit: nvm just saw your latest comment

PR #283, opened 2019-02-20
[Lou mocks] Header part 1

melissachang:
header

### DIFF
--- a/ui/public/index.html
+++ b/ui/public/index.html
@@ -1,16 +1,22 @@
 <!DOCTYPE html>
 <html lang="en">
   <head>
-    <meta charset="utf-8">
-    <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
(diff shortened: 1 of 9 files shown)


ui/public/index.html
```diff
@@ -1,16 +1,22 @@
 <!DOCTYPE html>
 <html lang="en">
   <head>
-    <meta charset="utf-8">
-    <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
-    <meta name="theme-color" content="#000000">
+    <meta charset="utf-8" />
```
wnojopra: The changes in this file seem to be just whitespace changes, is that intentional?

melissachang: In that file, I changed font from Lato to Montserrat. husky must have changed the rest. (I guess I didn't have husky set up in the earlier versions of that file.)

I could do a PR with just husky changes, let me know if you want that.

PR #285, opened 2019-02-26
[Lou mocks] Search box and drop-down

melissachang:


ui/src/components/Search.js
```diff
@@ -1,39 +1,117 @@
...
+    height: "31",
+    margin: "0 15px 0 0"
+  }),
+  multiValueLabel: (provided, state) => ({
+    ...provided,
+    color: "white",
+    fontSize: 12,
+    fontWeight: 500,
+    padding: "8px 21px 0 20px",
+    // For some reason paddingLeft from above is ignored; put again here
```
wnojopra: I wonder if you would experience this issue with other browsers?

melissachang: Not sure, and also not sure what you're getting at...?

PR #288, opened 2019-02-28
Never export samples/sample sets. Add dataset_name column to entities. Show export dialog even with no values selected.

wnojopra:
screenshot from 2019-02-28 12-25-17


wnojopra:
Got rid of the checkbox. Now we just never export samples/sample sets.

Also added the dataset_id: 
screenshot from 2019-02-28 13-53-21

Also fixed the dialog when there is no facets selected:
https://screencast.googleplex.com/cast/NDY3MjAwMjE3ODY3ODc4NHxlNTEyNDA4OC1kZA

melissachang:
Fixes #289

api/data_explorer/controllers/export_url_controller.py
```diff
@@ -113,41 +113,22 @@ def _get_entities_dict(cohort_name, query, filter_arr):
...
+                'dataset_id': table_name.split('.')[1],
             }
         })
 
-    # If a cohort was selected, create a query entity and get Elasticsearch documents
-    # for the cohort so we can create sample_set entity.
+    # If a cohort was selected, create a query entity
     if cohort_name:
         entities.append({
             'entityType': 'cohort',
```
melissachang: Right now there's no SQL query for the all_participants cohort. Let's do:
- Rename _get_filter_query -> _get_sql_query
- In _get_sql_query, if filter is empty, pick any participant table and return participant_ids from it
- Add TODO like

```
TODO: _get_sql_query() assumes that all tables have all participant ids.
Make _get_sql_query() work if that's not the case. participant ids should be
aggregated from all tables.
```

api/data_explorer/controllers/export_url_controller.py
```diff
@@ -113,41 +113,22 @@ def _get_entities_dict(cohort_name, query, filter_arr):
...
             'attributes': {
-                'table_name': table_name
+                'table_name': table_name,
+                'dataset_id': table_name.split('.')[1],
             }
         })
 
-    # If a cohort was selected, create a query entity and get Elasticsearch documents
-    # for the cohort so we can create sample_set entity.
+    # If a cohort was selected, create a query entity
```
melissachang: Delete 121-122

api/data_explorer/controllers/export_url_controller.py
```diff
@@ -113,41 +113,22 @@ def _get_entities_dict(cohort_name, query, filter_arr):
             # underscores here and periods in table_name attribute.
```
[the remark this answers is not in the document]
wnojopra: Can project ids have periods? "google.com:api-project-123"

And why are we splitting on the colon? Based on your description it sounds like we should be splitting on '.' instead.

I think what we actually want to do `'name': table_name.split('.')[-1]

wnojopra:
@melissachang I made the updates, lmk what you think.

PR #296, opened 2019-03-08
[Lou mocks] TextFacet checkbox

melissachang:
I started using [svg React components](https://facebook.github.io/create-react-app/docs/adding-images-fonts-and-files#adding-svgs). There is no need for custom Clarity icons; React components can do everything.

Also fix problem with there are lots of chips.

Before:

After:


ui/src/App.js
```diff
@@ -68,7 +68,7 @@ class App extends Component {
       // react-select default of "Select..." is shown.
       searchPlaceholderText: "",
       // Whether to show visualization facets or text facets
-      showViz: true,
+      showViz: false,
```
wnojopra: Why are we setting this to false? I liked seeing the visualizations by default.

melissachang: Fixed

PR #298, opened 2019-03-19
[Lou mocks] HistogramFacet part 1

melissachang:
I had to turn off tabular-nums for now. I will see if I can turn it back on in the next PR.

With tabular-nums, text is cut off:

Without tabular-nums, text is no longer cut off:

### DIFF
--- a/ui/src/components/facets/HistogramFacet.js
+++ b/ui/src/components/facets/HistogramFacet.js
@@ -1,7 +1,8 @@
 import React, { Component } from "react";
-import { withStyles } from "@material-ui/core/styles";
-import VegaLite from "react-vega-lite";
+import { canvas } from "vega-canvas";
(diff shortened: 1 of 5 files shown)


ui/src/components/facets/HistogramFacet.js
```diff
@@ -20,19 +22,34 @@ const styles = {
...
+  config: {
+    axis: {
+      labelColor: colors.gray[0],
+      labelFont: "Montserrat",
+      labelFontWeight: 500,
+      labelPadding: 9,
+      ticks: false
+    },
+    view: {
+      //      width: 250,
```
wnojopra: Is this needed?

melissachang: Done

PR #308, opened 2019-03-29
Change search controller regex filter

### LINKED ISSUE #297: Search strings should also work for variable substrings
Nicole brought this up - in Baseline searching for 'urine' doesn't match variables like standard_labs_albumin_urine_mgl or standard_labs_creatinine_urine_mgdl. Another example is in AMP PD, searching for 'territory' doesn't match genome_territory. 

A quick workaround while this feature doesn't exist is to add the substrings to the column descriptions.

wnojopra:
Together with https://github.com/DataBiosphere/data-explorer-indexers/pull/135 Fixes #297

### DIFF
--- a/api/data_explorer/controllers/search_controller.py
+++ b/api/data_explorer/controllers/search_controller.py
@@ -110,7 +110,7 @@ def search_get(query=None):
         response_fields = response.to_dict()
 
         # This regex matches if there is a word that starts with query,
-        # or a word ending with underscore_query.
+        # or a word that contains underscore_query.
         query_regex = r"\b{query}\w*|_{query}\w*".format(
             query=re.escape(query.lower()))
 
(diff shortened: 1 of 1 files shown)


api/data_explorer/controllers/search_controller.py
```diff
@@ -109,8 +109,10 @@ def search_get(query=None):
         current_app.logger.info("Elasticsearch query: %s sec" % (end - begin))
         response_fields = response.to_dict()
 
-        # This regex matches if there is a word that starts with query
-        query_regex = r"\b" + re.escape(query.lower()) + "\w*"
+        # This regex matches if there is a word that starts with query,
+        # or a word ending with underscore_query.
```
melissachang: `# or a word that contains underscore_query`

PR #318, opened 2019-04-11
Send data_explorer_url to Terra

wnojopra:
Just adds a new column to the data tab:
data_explorer_url


wnojopra:
Updated api.yaml, and ran the codegen scripts. Good to go.

PR #321, opened 2019-04-12
Propagate wid url param to import-data

melissachang:
If it exists, propagate wid url param to import-data url

- From Terra Data tab, user clicks on a button to view a cohort in DE. For DE url, terra-ui will append wid=`<wid>`, where `<wid>` is cohort's workspace id.
- In DE, user edits cohort and saves the new cohort. When DE calls app.terra.bio/#import-data, DE will pass wid param. The specified wid is selected by default in workspace drop-down.

Also:
- Change window.history.pushState -> replaceState
  I figure that if someone clicks around in DE and presses back button, they would want to go to page before DE

- Use new [UrlSearchParams library](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) to simplify query string parsing


ui/src/App.js
```diff
@@ -417,47 +417,35 @@ class App extends Component {
...
-
-  updateQueryParams(facetFilterParam, extraFacetsParam) {
-    window.history.pushState(
-      null,
-      "",
-      "?filter=" + facetFilterParam + "&extraFacets=" + extraFacetsParam
-    );
+  updateQueryString(filterParam, extraFacetsParam) {
+    // If there are any params other than filter/extraFacets, leave as-is.
+    var params = new URLSearchParams(window.location.search);
```
wnojopra: Awesome, this actually fixes problems with facet values containing '=' or '&'. We may still have to watch for the '|' character though

ui/src/App.js
```diff
@@ -417,47 +417,35 @@ class App extends Component {
...
+    params.set("filter", filterParam);
+    params.set("extraFacets", extraFacetsParam);
+    window.history.replaceState(null, "", "?" + params.toString());
```
wnojopra: Whats the reason to use replaceState? Using replaceState over pushState means we lose the ability to use the browser back/forward buttons (granted its still not working atm...)

melissachang: Discussed offline. At first I thought replaceState was preferable from a user's POV. I guess I could go either way -- it would be interesting to see what real users want.

Leaving at replaceState for now, since pushState doesn't work and fixing it appears to be non-trivial.

PR #322, opened 2019-04-19
Fix query string handling

melissachang:
I think #321 broke things. I thought I tested it locally, I'm not sure what happened.

Repro:
- Go to AMP PD DE and select a bar
- Copy url and open in new tab. You get a blank page

I tested this PR locally. I'm deploying now; I'll test on prod soon.

### DIFF
--- a/ui/src/App.js
+++ b/ui/src/App.js
@@ -405,7 +405,7 @@ class App extends Component {
   filterArrayToMap(filterArray) {
     let filterMap = new Map();
     filterArray.forEach(function(pair) {
-      pair = pair.split("%3D");
+      pair = pair.split(encodeURIComponent("="));
       if (filterMap.has(pair[0])) {
         let arr = filterMap.get(pair[0]);
         arr.push(pair[1]);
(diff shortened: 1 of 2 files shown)


wnojopra:
What do you think about adding a simple e2e test for query parameters? A barebones one might look like:
1) Click on a facet, check the url is updated as expected
2) goto a url with some query parameters. Make sure the page looks as expected.

In retrospect I probably should have added these tests when I first implemented.

melissachang:
Excellent idea, done. I confirmed test fails on master and works on this branch

ui/src/App.js
```diff
@@ -398,13 +399,13 @@ class App extends Component {
...
    * Example:
-   * In: ["Gender=female", "Gender=male", "Population=American"]
+   * In: ["Gender%3Dfemale", "Gender%3Dmale", "Population%3DAmerican"]
    * Out: {"Gender" => ["male", "female"], "Population" => ["American"]}
    */
   filterArrayToMap(filterArray) {
     let filterMap = new Map();
     filterArray.forEach(function(pair) {
-      pair = pair.split("=");
+      pair = pair.split("%3D");
```
wnojopra: Instead of hardcoding %3D, how about using `encodeURIComponent('=')` instead? I can never remember what these strings actually are. 

(Same with the %7C strings below)

melissachang: Good idea, done

PR #324, opened 2019-04-25
Remove TextFacet

melissachang:
This simplifies the UI by removing the viz switch, which will help when we embed DE into Terra. (For example, we won't be able to put viz switch in Terra header.)

AFAIK noone used TextFacet. If people request it, we can put it back later. Now I think it's better to add an arrow per facet, so each facet can be individually toggled.


ui/tests/e2e/test.js
```diff
@@ -11,219 +11,155 @@ describe("End-to-end", () => {
   beforeAll(async () => {
     // It can take a while for servers to start up
     jest.setTimeout(JEST_TIMEOUT_MS);
+    jest.setTimeout(10000000);
```
wnojopra: Is this necessary? Why?

melissachang: That was there for debugging, good catch. Deleted.

PR #323, opened 2019-04-20
Redesign in preparation for embedding Data Explorer inside Terra

melissachang:
This PR has been deployed to https://test-data-explorer.appspot.com
Move participant count out of header.
Change export fab to Save cohort button, to match Terra style.
In common.js, create buttons and tooltip that can be used throughout DE.

I deleted TextFacet/viz switch to simplify the UI. e2e tests fail because TextFacet no longer exists. I'll create another PR that deletes TextFacet and converts e2e tests to use HistogramFacet, and merge that before this PR. I wanted to send out this PR for review so you could look at the code.

PR #326, opened 2019-04-25
Copy URL button

### LINKED ISSUE #325: Add a 'copy-url' or 'share' button
Now that we have query parameters in the URL, users will be pushed into copying the URL to share to others. Sometimes users have issue with copying from the URL bar, so we should add a button to do it for them.

I'm unsure how UI/UX would look like for this, maybe browse around and look for good examples first.

wnojopra:
Fixes #325 
Depends on #323 (Should probably merge mc/move-out-header before reviewing this.

Screenshot from 2019-04-25 11-55-04
Clicking will put the URL into your clipboard.


ui/src/components/CopyUrlButton.js
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  handleButtonClick() {
+    // Just change the button text for a second
+    this.setState({ buttonText: "Copied!" });
+    setTimeout(() => this.setState({ buttonText: "Copy URL" }), 1000);
+  }
+
+  render() {
+    const { className } = this.props;
```
melissachang: className isn't used, delete 21 and 24

wnojopra: I think it used on 24, the styling is passed in through it.

wnojopra:
I had a lot of conflicts to fix, so decided it was easier to just open up https://github.com/DataBiosphere/data-explorer/pull/327 instead.

PR #327, opened 2019-04-26
Copy URL button

### LINKED ISSUE #325: Add a 'copy-url' or 'share' button

wnojopra:
Fixes #325

### DIFF
--- a/ui/src/components/CopyUrlButton.js
+++ b/ui/src/components/CopyUrlButton.js
@@ -3,12 +3,6 @@ import { CopyToClipboard } from "react-copy-to-clipboard";
 import { PrimaryButton, TerraTooltip } from "libs/common";
 import "@clr/icons/clr-icons.css";
 
-const style = {
-  copyUrlButton: {
-    marginRight: "16px"
-  }
-};
-
 export default class CopyUrlButton extends React.Component {
   constructor(props) {
     super(props);
@@ -31,7 +25,7 @@ export default class CopyUrlButton extends React.Component {
     const { className } = this.props;
 
     return (
-      <div onClick={this.handleButtonClick} style={style.copyUrlButton}>
+      <div onClick={this.handleButtonClick}>
         <TerraTooltip title={this.state.buttonText}>
           <CopyToClipboard text={window.location.href}>
             <clr-icon shape="copy-to-clipboard" size="20" />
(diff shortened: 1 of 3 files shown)


ui/src/components/CopyUrlButton.js
```diff
@@ -0,0 +1,43 @@
+import React from "react";
+import { CopyToClipboard } from "react-copy-to-clipboard";
+import { PrimaryButton, TerraTooltip } from "libs/common";
+import "@clr/icons/clr-icons.css";
+
+const style = {
+  copyUrlButton: {
+    marginRight: "16px"
```
melissachang: [React will automatically append a "px" suffix](https://reactjs.org/docs/dom-elements.html) so by convention we do this, to be slightly cleaner:
```
marginRight: 16
```

ui/src/components/CopyUrlButton.js
```diff
@@ -0,0 +1,43 @@
...
+      () => this.setState({ buttonText: "Copy URL to clipboard" }),
+      1000
+    );
+  }
+
+  render() {
+    const { className } = this.props;
+
+    return (
+      <div onClick={this.handleButtonClick} style={style.copyUrlButton}>
```
melissachang: It's [recommended to use className rather than style](https://reactjs.org/docs/dom-elements.html#style).

In this repo we use the [material-ui styling solution](https://material-ui.com/customization/overrides/#1-specific-variation-for-a-one-time-situation).
- At the bottom of this file: `export default withStyles(styles)(CopyUrlButton);`
-
```
const { classes } = this.props;
... className={classes.copyUrlButton}
```

melissachang (commented):
Sorry, can you move the copy button between search box and participant count? The Save button is important; this removes the noise from around it.

ui/src/components/CopyUrlButton.js
wnojopra: Good to know, thanks! In this case, since we wanted to move the copyUrl button to the left we don't even need the extra styling. So I just removed this altogether.

melissachang (dismissed):
I meant between search box and participant count. I pushed a commit with the changes (and some minor styling changes), let me know what you think.

melissachang (commented):
Actually let me get Sean's opinion, let me remove approval

PR #328, opened 2019-04-30
Changes for embedding DE into Terra Datasets library

melissachang:
https://i.imgur.com/uDjlVBA.mp4

Corresponds to https://github.com/DataBiosphere/terra-ui/pull/1533

### DIFF
--- a/ui/public/index.html
+++ b/ui/public/index.html
@@ -23,6 +23,12 @@
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
     <title>Data explorer</title>
+    <!--
+      Used for embedding Data Explorer into Terra.
+      Without this, Terra header is fixed: https://i.imgur.com/tucJzcG.gifv
+      With this, DE iframe embeds naturally: https://i.imgur.com/iqgwp8q.gifv
+    -->
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/iframe-resizer/4.1.1/iframeResizer.contentWindow.min.js"></script>
   </head>
   <body>
     <noscript> You need to enable JavaScript to run this app. </noscript>
--- a/ui/src/components/SaveButton.js
+++ b/ui/src/components/SaveButton.js
@@ -68,13 +68,18 @@ class SaveButton extends React.Component {
       <div className={className}>
         <TerraTooltip
           classes={{ tooltip: classes.tooltip }}
-          title="Save cohorts so you can work with them later"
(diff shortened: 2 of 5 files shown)


ui/public/index.html
```diff
@@ -23,6 +23,12 @@
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
     <title>Data explorer</title>
+    <!--
+      Used for embedding Data Explorer into Terra.
+      Without this, Terra header is fixed: https://i.imgur.com/tucJzcG.gifv
+      With this, DE iframe embeds naturally: https://i.imgur.com/iqgwp8q.gifv
+    -->
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/iframe-resizer/4.1.1/iframeResizer.contentWindow.min.js"></script>
```
wnojopra: I think it might be worth adding this comment from the github's readme:

> The second file (iframeResizer.contentWindow.min.js) needs placing in the page(s) contained within your iFrame. This file is designed to be a guest on someone else's system, so has no dependencies and won't do anything until it's activated by a message from the containing page.

ui/src/components/SaveButton.js
```diff
@@ -152,28 +157,48 @@ class SaveButton extends React.Component {
       if (error) {
         alert(error.response.body.detail);
       } else {
-        let importUrl =
-          "https://app.terra.bio/#import-data?format=entitiesJson";
+        let queryStr = "format=entitiesJson";
```
wnojopra: I feel like we shouldn't be manually appending strings together when URLSearchParams exist. I think we should change this to:
`let queryParams = new URLSearchParams("format=entitiesJson")`

And then when adding to it, do (ex):
`queryParams.append("wid", wid)`

And then when calling sendMessage, do:
`queryParams.toString()`

wnojopra:
If I'm in Terra looking at an embedded data explorer, and I copy the URL in the browser, is this something I can share?

What if instead I click on a copy url button, is that any different?

ui/public/index.html
melissachang: I think it's ok, if people are curious they can google and find that

ui/src/components/SaveButton.js
melissachang: IMHO we should use URLSearchParams when it makes the code more readable/better. In this case, I think it actually makes the code a little less readable -- since the code is quite clear on its own.

PR #330, opened 2019-05-01
Fix javascript warnings

melissachang:
Also configure CircleCI to fail if there are JS warnings, so new warnings don't creep in.

When DE is embedded in Terra, DE warnings will appear in Terra console (maybe?).


melissachang:
We use create-react-app which uses react-scripts. The default behavior is:
- For local runs (npm start), don't fail on js warnings
- For prod builds (npm build), [fail on js warnings only if CI=true.](https://github.com/facebook/create-react-app/blob/b36d1ea1aa1458907839da9dd5028a372f9d48d0/packages/react-scripts/scripts/build.js#L186)

`npm run build` is used to make prod deployment build. (It minifies files, etc) Most projects have automated deployment and run `npm run build` in some CI environment (circleci, travis) -- eg see [terra-ui](https://github.com/DataBiosphere/terra-ui/blob/9e6f07f5bf6b4643df4a565e8ae5f57f8b23e908/.circleci/config.yml#L67). Those environments set `CI=true`.

We are still in stone ages and manually deploy with `gcloud app deploy`. (We run [`npm run build`](https://github.com/DataBiosphere/data-explorer/blob/eae86a652ce5120a07877177c2c325245d286b5f/ui/Dockerfile#L17) but `CI` isn't set to `true`.)

I want to fail before merge, rather then before deployment, so I changed circleci rather than [our `npm run build`](https://github.com/DataBiosphere/data-explorer/blob/master/ui/Dockerfile#L17).

PR #333, opened 2019-05-16
Preliminary load testing

melissachang:
Requested by @seanhorgan.

I simulated 20-100 simultaneous users of NHS Data Explorer, to simulate a workshop. Data Explorer was slower, but still usable.

The slowest part of Data Explorer is doing a search on large datasets like NHS. Searching for `pre` takes ~6 sec; `men` takes ~5s.

**20 simultaneous searches**
- DE loaded in all tabs
- Search times: 4 3 8 7 11 10 14 12 16 15 19 19 20 22 23 25 25 27 27 29

**100 simultaneous searches**
- DE didn't load for 35 out of 100 tabs. (To me this is fine; user can just reload.)
- For tabs where DE did load, search times: 4 4 8 6 9 13 8 15 10 13 18 16 20 18 23 19 26 21 30 18 32 21 33 26 34 27 36 32 42 16 41 37 48 40 46 37 49 47 57 46 61 38 59 36 66 46 64 45 71 47 61 48 72 31 73 55 76 36 61 59 71 35 75 33 68 41

**Implementation notes**

Unfortunately this requires manually scrolling through Chrome tabs.
- When headless if off, [only the focussed tab browser test code will execute](https://github.com/GoogleChrome/puppeteer/issues/2254).
- This issue doesn't happen when headless is on. But when headless is on, I'm not able to manually login to channing account. I tried writing code to automate this, but got stuck because Google starting asking for captcha.
- [Multiple browser contexts](https://github.com/GoogleChrome/puppeteer/issues/4219#issuecomment-479224433) don't have this problem -- With headless off, test code in all contexts runs simultaneously. However, I'd have to manually log into each context which is a pain.

### DIFF
--- a/ui/tests/load/test.js
+++ b/ui/tests/load/test.js
@@ -1,3 +1,18 @@
+/*
+
+Simple load test. Does 100 simultaneous searches on NHS.
+Use NHS because it is a large dataset. (UKBB is also large, but is mostly
+numbers. NHS has strings and so is more realistic.)
+
+- Uncomment headless in jest-puppeteer.config.js
+- cd ui && ./node_modules/.bin/jest tests/load
+- In test window that popped up, login as channing user
+- In test window, Ctrl-Pgdown to scroll through all tabs. (See XXX for why
+this manual step is needed.) Each tab (except for first) should say "Loading..."
+in search box. This means search has started.
+
+*/
+
 const JEST_TIMEOUT_MS = 600 * 1000;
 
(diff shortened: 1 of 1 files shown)


melissachang:
Simultaneous searching would only happen in the context of a workshop, so I'm not too concerned about the results. For workshop, I just want to make sure the application doesn't totally fall over.

For NHS, we get about 1-2 users a week. It would be a long time before we get many simultaneous searches organically.

PR #337, opened 2019-05-28
Fix all participants query

melissachang:
Instead of getting participant ids from first table, get from all tables. Tested with 1000 Genomes and AMP PD. Before this PR, AMP PD all participants query returned 5407 participants. After this PR, 5478 partcipants are returned.

Also some code clean up. Follow this repo's convention of putting helper methods before they're used.

### DIFF
--- a/api/data_explorer/controllers/export_url_controller.py
+++ b/api/data_explorer/controllers/export_url_controller.py
@@ -133,6 +133,18 @@ def _get_table_and_clause(es_field_name, field_type, value,
     return table_name, column, clause
 
 
+def _get_all_participants_query():
+    participant_id = current_app.config['PARTICIPANT_ID_COLUMN']
+    query = 'SELECT DISTINCT %s FROM (' % participant_id
+    query += 'SELECT %s FROM `%s`' % (participant_id,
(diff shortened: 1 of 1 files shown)


api/data_explorer/controllers/export_url_controller.py
```diff
@@ -336,28 +244,89 @@ def _append_to_query(existing, new, join, table_num):
...
+    current_app.logger.info('Signed URL: ' + signed_url)
+    return signed_url
 
-    entities = _get_entities_dict(cohort_name, query, filter_arr,
-                                  data_explorer_url)
 
-    # Don't actually write GCS file during unit test. If we wrote a file during
-    # unit test, in order to make it easy for anyone to run this test, we would
-    # have to create a world-readable bucket.
-    if 'pytest' in sys.modules:
```
wnojopra: Why is this being removed? Did we remove the relevant unit tests? I forget.

melissachang: Yes #262 

Only test left is https://github.com/DataBiosphere/data-explorer/blob/master/api/data_explorer/test/test_elasticsearch_util.py

PR #338, opened 2019-05-28
New Terra colors

melissachang:
Corresponds to https://github.com/DataBiosphere/terra-ui/pull/1621


ui/src/App.js
```diff
@@ -23,16 +22,16 @@ import Montserrat from "libs/fonts/Montserrat-Medium.woff2";
 
 const styles = {
   disclaimer: {
-    color: colors.gray[3],
+    color: "#333f52",
```
wnojopra: How come we hard code in several places? Can we do colors.dark here (for example)?

melissachang: I did the ones that were easy to fix. For some colors, it's hard to tell what the `color.*()` version is. Note that having `color.*()` doesn't necessarily make the code easier to maintain -- we had that prior to this PR and it actually made updating the colors harder.

PR #340, opened 2019-05-31
Allow all origins for CORS on export bucket

melissachang:
Export was failing for firecloud.terra.bio, and various test instances.


melissachang:
Fixed, thanks

melissachang:
Oops, I forgot CORS was necessary since Terra is fetching from client. Changed to all origins

PR #341, opened 2019-07-16
Add time series support to Data Explorer

lgolowich:
@wnojopra


api/data_explorer/__main__.py
```diff
@@ -199,26 +207,43 @@ def _process_facets(es):
...
-        }
-        if 'ui_facet_description' in facet_config:
-            facets[es_field_name]['description'] = facet_config[
-                'ui_facet_description']
-
-        facets[es_field_name][
-            'es_facet'] = elasticsearch_util.get_elasticsearch_facet(
-                es, es_field_name, field_type)
+        es_base_field_name = facet_config['elasticsearch_field_name']
+        if elasticsearch_util.is_time_series(es, es_base_field_name):
```
wnojopra: I would move the time_series specific code to its own new function.

Something like

is_time_series, time_series_vals, es_field_names = get_time_series_info()

api/data_explorer/controllers/export_url_controller.py
```diff
@@ -77,31 +78,31 @@ def _check_preconditions():
         raise BadRequest(error_msg)
 
 
-def _get_range_clause(column, value):
+def _get_range_clause(column, value, bucket_interval):
     arr = value.split('-')
     if len(arr) > 1:
         low = arr[0]
         high = arr[1]
```
wnojopra: I'm not exactly sure what we're going for here, but based on the code below it looks like we're not using this 'high' variable. Lets delete it if so.

api/data_explorer/controllers/export_url_controller.py
```diff
@@ -166,13 +190,24 @@ def _get_sql_query(filters):
         facet_id = splits[0]
         value = splits[1]
         field_type = ''
+        is_time_series = False
```
wnojopra: Seems like is_time_series and bucket_interval will always be defined in one of the if statements below. So not necessary to declare these here.

api/data_explorer/controllers/facets_controller.py
```diff
@@ -20,32 +19,52 @@ def _get_bucket_interval(facet):
...
-            'ui_facet_name': ui_facet_name,
-            'type': field_type
-        }
-        facets[elasticsearch_field_name][
-            'description'] = elasticsearch_util.get_field_description(
-                es, elasticsearch_field_name)
-        facets[elasticsearch_field_name][
-            'es_facet'] = elasticsearch_util.get_elasticsearch_facet(
-                es, elasticsearch_field_name, field_type)
+        if elasticsearch_util.is_time_series(es, es_base_field_name):
```
wnojopra: I would write this as 

is_time_series = elasticsearch_util.is_time_series(es, es_base_field_name)
if is_time_series:
    ....

.gitignore
```diff
@@ -2,8 +2,11 @@
 dataset_config/*
 !dataset_config/template
 !dataset_config/1000_genomes
+!dataset_config/framingham_heart_study_teaching
```
melissachang: Nit let's put all the framingham lines together, under the 1000_genomes lines

api/api.yaml
```diff
@@ -124,6 +124,11 @@ definitions:
       search_placeholder_text:
         type: string
         description: Optional. What to show in the search box by default
+      time_series_unit:
+        type: string
+        description: >
+          Optional. The unit of time (such as month, year, etc.) with
```
melissachang: 130-131:

If time_series_column in bigquery.json is set, this must be set. This is used to label the time axis -- eg Month or Year.

FYI search_placeholder_text is optional because if it's not set here, the UI will show something by default

api/api.yaml
```diff
@@ -153,19 +158,36 @@ definitions:
...
+        type: array
+        items:
+          type: array
+          items:
+            type: integer
+        description: >
+          2-dimensional array of facet value counts, indexed by time
+          then value; indexes correspond to time_names and then
+          value_names.
+      is_time_series:
```
melissachang: What do you think about deleting this and checking if time_names or time_series_value_counts is set instead? I think that's cleaner.

For Elasticsearch, IIRC there wasn't an easy way to tell if a field was time series; that's why we added _is_time_series field

api/api.yaml
```diff
@@ -153,19 +158,36 @@ definitions:
       es_field_type:
         type: string
         description: The Elasticsearch field type.
-      values:
+      value_names:
```
melissachang: Delete 166-167

Put on 161:

If this is a regular facet (https://i.imgur.com/ug1mSEr.png), only value_names and value_counts are set.
If this is a time series facet (https://i.imgur.com/IU2SqWv.png), then value_names, time_names, time_series_value_counts are set. Consider https://i.imgur.com/IU2SqWv.png: the data is a table with 6 rows and 3 columns. Column names are time_names; row names are value_names; time_series_value_counts are the numbers in the table.

api/api.yaml
```diff
@@ -199,3 +221,8 @@ definitions:
         description: >
           If this search result represents a facet, facet_value is the empty string.
           If this search result represents a facet and selected value, facet_value is the selected value.
+      is_time_series:
```
melissachang: Why is this needed? (I haven't read the code yet)

ui/src/components/Search.js
```diff
@@ -145,14 +145,30 @@ class Search extends React.Component {
...
     if (option.facetValue !== null && option.facetValue !== "") {
-      return (
-        <div>
-          <span style={{ color: "#cccfd4" }}>Add</span>
-          <span> {option.facetName} </span>
-          <span style={{ color: "#cccfd4" }}>facet and select</span>
-          <span> {option.facetValue} </span>
-        </div>
-      );
+      if (option.isTimeSeries) {
```
melissachang: time series could have description

ui/src/components/facets/HistogramPlot.js
```diff
@@ -0,0 +1,300 @@
+import React, { Component } from "react";
```
melissachang: nit can you rename from HistogramPlot.js to Histogram.js

ui/src/components/facets/TimeSeriesHistogramPlot.js
```diff
@@ -0,0 +1,285 @@
+import React, { Component } from "react";
```
melissachang: I assume this is multiple histograms. Rename this file to TimeSeriesHistograms.js (plural).

Please add a comment and you can reference https://i.imgur.com/YsAV5P6.png

api/api.yaml
lgolowich: The only case where is_time_series is needed here is when the facet has no data, so that value_names, value_counts, time_names, and time_series_value_counts are all empty. The current implementation of empty time series facets doesn't show the zero count in the upper right corner, whereas non-time series facets do. (The heatmap also had empty time series facets be twice as wide). Do you think it's cleaner just to get rid of these differences, or else explain this in the is_time_series description?

melissachang: I prefer to not have `is_time_series` here. I don't think we even need to mention the "no data" case in these comments... for time series facets with no data, I don't have a strong opinion on whether we should show zero in upper right or not.

lgolowich: Ok, will remove.

api/api.yaml
lgolowich: It's used to determine when to add the "at {time series unit} {time}" clause to the search option. Should this go in the description?

melissachang: Ah ok. This doesn't need to go in description. But please make boolean instead of integer.

(The logic in Search.js is getting complicated. In general I prefer to have complicated logic in API server and not UI server. So in retrospect it might be better to have API server just return es_field_name and the html of what to render. Probably not worth changing at this point.)

api/data_explorer/__main__.py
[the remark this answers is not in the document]
lgolowich: Yes, and I think using `mapping` is also simpler for `get_field_type`; I'll update these functions.

api/data_explorer/__main__.py
```diff
@@ -199,26 +207,43 @@ def _process_facets(es):
...
+            facets[es_field_name] = {
+                'ui_facet_name': ui_facet_name,
+                'type': field_type,
+                'is_time_series': is_time_series
+            }
+            if 'ui_facet_description' in facet_config:
+                facets[es_field_name]['description'] = facet_config[
+                    'ui_facet_description']
+
+            if field_type != 'text' and field_type != 'boolean' and interval < 0:
```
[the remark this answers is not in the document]
lgolowich: The one issue with calling get_bucket_interval in get_elasticsearch_facet in the current setup is that for time series facets, get_bucket_interval (which makes a call to elasticsearch) would then be called redundantly for each possible time, which could be inefficient. Maybe it would be cleaner to have a wrapper function get_time_series_elasticsearch_facets to deal with this?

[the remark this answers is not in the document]
melissachang: Can you try the readable/inefficient approach and see if it's slower? If you don't notice a performance difference, I prefer more readable. (Or if you can think of a readable, efficient approach, that would be fine too.)

Also FYI, the idiomatic Python way of `interval = -1` is `interval = None`/`if not  interval`.

api/data_explorer/controllers/facets_controller.py
```diff
@@ -84,30 +120,112 @@ def facets_get(filter=None, extraFacets=None):  # noqa: E501
...
+    ts_description = ""
+    ts_field_type = ""
+    ts_time_names = []
+    ts_value_names = {}
+    ts_values = []
+
     for es_field_name, facet_info in combined_facets.iteritems():
-        values = []
+        if ts_field_name and (
+            (not facet_info.get('is_time_series'))
```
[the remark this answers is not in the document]
lgolowich: combined_facets contains one "facet" for each bottom level elasticsearch field, which means a time series facet with 3 separate times is stored as 3 consecutive elements of combined_facets. This code (and the similar code at the end of the function) is used to package together these separate "facets" into one Facet object representing a time series facet.

I'm doing what you suggested below and reorganized this function into helper functions, so I think the new version will be more clear.

api/data_explorer/controllers/facets_controller.py
```diff
@@ -20,32 +19,52 @@ def _get_bucket_interval(facet):
...
         return
 
-    for elasticsearch_field_name in extra_facets:
-        if not elasticsearch_field_name:
+    es = Elasticsearch(current_app.config['ELASTICSEARCH_URL'])
+    facets = OrderedDict()
+    mapping = es.indices.get_mapping(index=current_app.config['INDEX_NAME'])
+
+    for es_base_field_name in extra_facets:
+        if not es_base_field_name:
```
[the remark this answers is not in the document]
lgolowich: It seems that this case can happen when filter is empty (such as with the default filter=&extraFacets=). I guess that should be documented in a comment?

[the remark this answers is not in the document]
melissachang: How about this, I think it's a little more clear.
```
if not extra_facets:
  return

for es_base_field_name in extra_facets:
        if elasticsearch_util.is_time_series(es, es_base_field_name):
            is_time_series = True
```

api/data_explorer/util/elasticsearch_util.py
```diff
@@ -235,6 +266,46 @@ def get_field_type(es, field_name):
...
+        return True
+    elif mapping:
+        return False
+    else:
+        raise ValueError(
+            'elasticsearch_field_name %s not found in Elasticsearch index %s' %
+            (field_name, current_app.config['INDEX_NAME']))
+
+
+def get_time_series_vals(es, field_name, mapping):
```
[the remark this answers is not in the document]
lgolowich: It's just a 1D array. I'll add "Returns a sorted array of the times at which field_name could have data."

ui/src/components/Search.js
lgolowich: The `else if (option.facetDescription != null)` statement is at a lower nesting level than the `if (option.isTimeSeries)` statement, so it should still apply to both time series and non time series fields, unless I'm missing a case?

api/data_explorer/controllers/facets_controller.py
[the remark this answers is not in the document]
lgolowich: Ok. Turns out extra_facets could also be [''], but I added this to the if statement.

ui/src/components/Search.js
melissachang: Ah ok

dataset_config/1000_genomes/bigquery.json
```diff
@@ -14,21 +14,30 @@
...
-        "WGS High Coverage CRAM":"verily-public-data.human_genome_variants.1000_genomes_sample_info.wgs_high_cov_cram",
-        "WGS Low Coverage CRAM":"verily-public-data.human_genome_variants.1000_genomes_sample_info.wgs_low_cov_mapped_cram",
-        "Exome BAM":"verily-public-data.human_genome_variants.1000_genomes_sample_info.exome_mapped_bam",
-        "Exome CRAM":"verily-public-data.human_genome_variants.1000_genomes_sample_info.exome_mapped_cram"
-   }
+// time_series_column:
+//   The (optional) name of BigQuery column that holds the time
+//   associated with a data point. This column must have a numeric
+//   type.
+// time_series_unit:
```
melissachang: Don't forget to add to other repo's config files

PR #345, opened 2019-07-30
Added floating point support

lgolowich:
@wnojopra


api/api.yaml
```diff
@@ -181,8 +181,11 @@ definitions:
       time_names:
         type: array
         items:
-          type: integer
-        description: Array of times.
+          type: string
```
melissachang: In general I like having business logic in API server instead of UI server. Have API convert _ to a float number, and change this type to float.

PR #348, opened 2019-08-05
Add legal disclaimer for Framingham

melissachang:
Also change button from `Save to Terra` back to `Save cohort`. (I tried exporting samples and ran into scalability issues.)

Also expand README a bit.

### DIFF
--- a/api/api.yaml
+++ b/api/api.yaml
@@ -180,6 +180,7 @@ definitions:
         description: Array of counts for each facet value.
       time_names:
         type: array
+        # This is string because time may be "Unknown"
         items:
           type: string
         description: Array of times.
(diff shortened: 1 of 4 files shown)


api/api.yaml
```diff
@@ -180,6 +180,7 @@ definitions:
         description: Array of counts for each facet value.
       time_names:
         type: array
+        # This is string because time may be "Unknown"
```
lgolowich: In the API I think it's technically "unknown" (not too important).

melissachang: Ah. In general the API server should pass what to render, UI server shouldn't have to make changes. How about we store in Elasticsearch as "Unknown"?

lgolowich: Sounds good, I'll make that change.

PR #350, opened 2019-08-07
Added support for separate time series panels

lgolowich:
@wnojopra


api/api.yaml
```diff
@@ -230,5 +230,5 @@ definitions:
       is_time_series:
         type: boolean
         description: >
-          Is true when facet_value is nonempty and the field is a time
-          series field.
+          Is true when elasticsearch_field_name gives a complete time
```
melissachang: True if this is a time series field at a specific point in time

(Is that right?)

api/data_explorer/__main__.py
```diff
@@ -212,38 +249,35 @@ def _process_facets(es):
...
-                'ui_facet_name': ui_facet_name,
-                'type': field_type,
-                'is_time_series': is_time_series
-            }
-            if 'ui_facet_description' in facet_config:
-                facets[es_field_name]['description'] = facet_config[
-                    'ui_facet_description']
-            facets[es_field_name][
-                'es_facet'] = elasticsearch_util.get_elasticsearch_facet(
-                    es, es_field_name, field_type, time_series_vals)
```
melissachang: Can you update 282? By convention we document the things in app.app.config somewhere in this method. Eg I think time_series_panel is entries in the dict.

api/api.yaml
lgolowich: Yes, will update

api/data_explorer/__main__.py
```diff
@@ -212,42 +249,44 @@ def _process_facets(es):
...
-
-    # Map from Elasticsearch field name to dict with ui facet name,
-    # Elasticsearch field type, optional UI facet description and Elasticsearch
-    # facet.
+            _add_facet(es_base_field_name, is_time_series, False, True,
+                       facet_config, [], facets, es, mapping)
+
+    # Map from Elasticsearch field name to dict with:
+    # - ui facet name
```
melissachang: For convenience, use dict key name (`ui_facet_name`)

api/data_explorer/__main__.py
```diff
@@ -212,42 +249,44 @@ def _process_facets(es):
...
+    # - Elasticsearch field type
+    # - time_series_panel and separate_panel, booleans indicating if the
```
melissachang:
```
# - time_series_panel: If facet is part of a time series facet. (The time series facet could be over all time, or at one point in time.)
# - separate_panel: For time series facets, if the time series histogram is in a separate panel (https://i.imgur.com/JsTb5r0.png), as opposed to part of the larger time series facet (https://i.imgur.com/7txytrL.png)
```

lgolowich:
Sorry I didn't explain it well before - I've tried rewording the descriptions in f82628ec3cc9f4914eaf5e929bb438b74fdbc1f0

PR #354, opened 2019-08-13
Fixed sorting of time series columns

lgolowich:
@wnojopra 

Although no problems had surfaced before, I got some when experimenting with a total column.


lgolowich:
For some reason I can only reproduce it by adding another column. For example, if [these lines](https://github.com/DataBiosphere/data-explorer/blob/ea86b54e2906778fb7a35b671bd8ca88a0e22750/ui/src/components/facets/TimeSeriesHistograms.js#L159-L160) are replaced by `type: "nominal"`, then the Total and Unknown columns are in the wrong order for me for the `biological_mother_with_pd` field.

PR #359, opened 2019-10-01
Return 10000 search results instead of 100

### LINKED ISSUE #292: Please increase the number of fields displayed in the search bar.
In the AMP PD data explorer, when I click on the down arrow or in the search bar itself, a list of facet fields pops up.

These only appear to be for the sample-related tables, not the participants.
It seems I need to know the participant fields to search on.

melissachang:
Fixes #292 

This is possible due to react-window.

I tried with local UI/API server talking to NHS ES, things seemed to work.

### DIFF
--- a/ui/src/components/Search.js
+++ b/ui/src/components/Search.js
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { Component } from "react";
 // custom-elements needed for old versions of Firefox. Must be before @clr.
 import "@webcomponents/custom-elements";
 import "@clr/icons";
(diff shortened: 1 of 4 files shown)


ui/src/components/Search.js
```diff
@@ -324,6 +325,29 @@ class Search extends React.Component {
       );
     };
 
+    // Use react-window to make UI faster when there are a lot of search
+    // results. See
+    // https://github.com/JedWatson/react-select/issues/2850#issuecomment-410318717
+    const height = 35;
```
wnojopra: Does this need to be out here? Shouldn't it go inside the MenuList's render method?

melissachang: Done

melissachang:
Done

PR #360, opened 2019-10-03
Handle invalid facets in url params in a user-friendly way

melissachang:
This PR addresses the scenario when a dataset version changes and cohorts no longer work. For example, AMP PD may release a new dataset quarterly.

- AMP PD v1 data is stored in BigQuery dataset `2019_v1_0101`.
- Data Explorer url contains `filter=amppd.2019_v1_0101.demographics.sex=female` which is valid. User saves a cohort with this filter.
- A new version of AMP PD is released. (Data Explorer URL remains the same.) Elasticsearch index is recreated with the new BigQuery dataset `2019_v2_0401`.
- User won't be able to open saved cohort in DE; `amppd.2019_v1_0101.demographics.sex` is no longer is Elasticsearch index.

Before this PR, if filter or extraFacets url params contained invalid facet names, DE UI is blank.

After this PR:
- DE renders, ignoring the invalid facets
- A snackbar is shown so user knows some of their filters/extra facets are missing
- Invalid facets are removed from url, so if user tries to saves cohort, cohort won't contain invalid params

Invalid filter param:

Multiple invalid filter params:

For invalid extra facets param, we show slightly different text:

Multiple invalid extra facets params:

Also tested:
- Invalid filter and extra facet. You see the invalid extra facet snackbar first. After you dismiss that, then you see the invalid filter snackbar.
- Invalid filter, valid extra facet
- Invalid extra facet, valid filter
- Invalid and valid filters. Only invalid filters are removed from url.
- Invalid and valid extra facets. Only invalid extra facets are removed from url.
- Embedded and standalone. When embedded, invalid params are removed from parent url in addition to iframe url
- Time series
  - Invalid filter in "all time periods" facet - filter correctly removed from url
  - Invalid filter in "single time period" facet - filter correctly removed from url
  - Invalid extra facet that is "all time periods" facet - extra facet correctly removed from url
  - Invalid extra facet that is "single time period" facet - extra facet correctly removed from url
  - Invalid filter in "single time period" facet and invalid extra facet that is "single time period" facet
- "How to use Data Explorer" snackbar still goes away second time


api/data_explorer/controllers/facets_controller.py
```diff
@@ -267,13 +270,15 @@ def facets_get(filter=None, extraFacets=None):  # noqa: E501
     :type extraFacets: List[str]
     :rtype: FacetsResponse
     """
-    _process_extra_facets(extraFacets)
+    es = Elasticsearch(current_app.config['ELASTICSEARCH_URL'])
+    invalid_extra_facets = _process_extra_facets(es, extraFacets)
```
wnojopra: It seems a little weird to me that _process_extra_facets also returns a list of invalid facets. How about we have a function that takes the extraFacets, and returns two lists: one valid and one invalid. Then we pass the valid ones to _process_extra_facets. I think this makes it more clear what the functions are meant to do.

api/data_explorer/controllers/facets_controller.py
```diff
@@ -267,13 +270,15 @@ def facets_get(filter=None, extraFacets=None):  # noqa: E501
...
     combined_facets = (current_app.config['EXTRA_FACET_INFO'].items() +
                        current_app.config['FACET_INFO'].items())
     combined_facets_dict = OrderedDict(combined_facets)
-    search = DatasetFacetedSearch(
-        elasticsearch_util.get_facet_value_dict(filter, combined_facets_dict),
-        combined_facets_dict)
+
+    filter_dict, invalid_filter_facets = elasticsearch_util.get_facet_value_dict(
```
wnojopra: My previous comment applies to this function as well.

api/data_explorer/util/elasticsearch_util.py
```diff
@@ -188,33 +188,59 @@ def convert_to_index_name(s):
...
+            filter_dict[es_field_name].append(facet_value)
+    return filter_dict, invalid_filter_facets
+
+
+def field_exists(es, field_name):
+    try:
+        es.get(index=current_app.config['FIELDS_INDEX_NAME'],
+               doc_type='type',
+               id=field_name)
+    except Exception:
```
wnojopra: In this function, can we be specific about what exception we catch here? We should try not to cover up any potential bugs further down the stack.

ui/src/components/DeHeader.js
```diff
@@ -68,70 +48,97 @@ const styles = {
...
+      invalidExtraFacets && invalidExtraFacets.length
+        ? createInvalidFacetsSnackbar(
+            classes,
+            invalidExtraFacets,
+            <div>
+              Try searching for
+              {invalidFilterFacets.length === 1
+                ? " this facet "
+                : " these facets "}
+              and saving cohort to Terra again.
```
wnojopra: Does "saving *the* cohort to Terra again" sound better?

api/data_explorer/controllers/facets_controller.py
melissachang: I don't think it's weird and I think it's cleaner this way. Have you ever programmed in Go? In Go, it's common to return errors along with valid results. 

This is different from languages like Java where you can only return one thing. Maybe it takes a little getting used to...

If you want, I can add pydoc to make it clear what is being returned.

api/data_explorer/util/elasticsearch_util.py
melissachang: Done

ui/src/components/DeHeader.js
melissachang: Done

api/data_explorer/controllers/facets_controller.py
wnojopra: Thanks, I see. I'd be happy with documentation on what is being returned.

melissachang: Done. This was a good idea anyways as other things also weren't clear.

PR #362, opened 2019-10-10
For extra facets, print dataset and table in description

melissachang:
This is useful for datasets with many BQ datasets and tables.

Tested on different extra facet types:
- Normal facet
- Samples facet
- Time series facet, all times
- Time series facet, single time


melissachang:
Here's a long description, I think it looks ok

PR #372, opened 2020-06-15
Remove 'Samples Overview' facet

wnojopra:
Tested with 1000 genomes and AMP PD.

### DIFF
--- a/api/data_explorer/controllers/export_url_controller.py
+++ b/api/data_explorer/controllers/export_url_controller.py
@@ -116,10 +116,6 @@ def _get_table_and_clause(es_field_name, field_type, value, bucket_interval,
     eg "((age76 >= 20 AND age76 < 30) OR (age76 >= 30 AND age76 < 40))".
     """
     sample_file_type_field = False
-    if field_type == 'samples_overview':
-        es_field_name = current_app.config['FACET_INFO']['Samples Overview'][
-            'elasticsearch_field_names'][value]
-        value = True
 
     if es_field_name.startswith('samples.'):
         es_field_name = es_field_name.replace('samples.', '')
--- a/api/data_explorer/controllers/export_url_controller.py
+++ b/api/data_explorer/controllers/export_url_controller.py
@@ -196,35 +196,33 @@ def _get_sql_query(filters):
     facet_table_clauses = {}
     for filter_str in filters:
         splits = filter_str.rsplit('=', 1)
-        # facet_id is "Samples Overview" for the Samples Overview facet, and
-        # es_field_name for all other facets.
-        facet_id = splits[0]
+        es_field_name = splits[0]
         value = splits[1]
         field_type = ''
         is_time_series = False
         bucket_interval = None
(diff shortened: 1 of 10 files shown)


wnojopra:
@melissachang Thanks, sorry I forgot to do that in the first place. Tested again with 1000 genomes and amp pd looks good.

api/data_explorer/controllers/export_url_controller.py
```diff
@@ -116,10 +116,6 @@ def _get_table_and_clause(es_field_name, field_type, value, bucket_interval,
     eg "((age76 >= 20 AND age76 < 30) OR (age76 >= 30 AND age76 < 40))".
     """
     sample_file_type_field = False
-    if field_type == 'samples_overview':
-        es_field_name = current_app.config['FACET_INFO']['Samples Overview'][
-            'elasticsearch_field_names'][value]
-        value = True
```
melissachang, line 122: The `_has_` fields on line 123 is only used for Samples Overview right?

If so, please delete logic around `_has_`. (These fields should be deleted in indexer repo as well.)

melissachang, line 122: Later on in this file:

```
        # facet_id is "Samples Overview" for the Samples Overview facet, and
        # es_field_name for all other facets.	        # es_field_name for all other facets.
```

Now that we no longer have "Samples Overview", get rid of `facet_id`. Just use `es_field_name`.

melissachang, line 122: There are other places in the file that mention "Samples Overview", please clean up all of them.

dataset_config/1000_genomes/bigquery.json
```diff
@@ -12,8 +12,7 @@
 //   The (optional) name of BigQuery column that holds the sample ID.
 // sample_file_columns:
 //   An (optional) map from file type to BigQuery column containing file paths.
```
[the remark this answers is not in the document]
wnojopra: For AMP PD, instead of the sample overview facet, we have separate facets for "Has WGS sample" and "Has RNA sample". We specify the columns here so that generating sql queries for these facets work properly.

api/data_explorer/controllers/export_url_controller.py
wnojopra: We're using the _has fields for AMP PD in their own facets, so I'd like to keep this here and in the indexer.

wnojopra: Maybe I'm misunderstanding what you're saying here, but es_field_name isn't in this function. Are you suggesting to rename the variable facet_id to es_field_name?

wnojopra: Cleaned up those comments.

melissachang: Yes. The only reason we called it `facet_id` instead of `es_field_name` was because of the Samples Overview facet.

Looks like you got most of the places, I think you missed a couple.

PR #373, opened 2020-06-16
Render true -> True and false -> False

wnojopra:
Tested with 1000 genomes (In Final Phase Variant Calling facet has boolean value) and AMP PD looks good.


wnojopra:
Wow, thanks @melissachang this makes a lot more sense. Testing looks good.

api/data_explorer/controllers/facets_controller.py
```diff
@@ -185,8 +185,10 @@ def _get_facet_values(es_field_name, facet_info, es_response_facets):
         else:
             # elasticsearch-dsl returns boolean field keys as 0/1. Use the
             # field's 'type' to convert back to boolean, if necessary.
+            # Also cast the bool to a str so that it renders as True/False
+            # instead of true/false (uppercase).
```
melissachang, line 189: opt nit - delete "(uppercase)"

PR #376, opened 2020-06-17
Add pre-flight check for valid private-key.json

### LINKED ISSUE #375: Check if private-key.json file is valid
We currently give an error if the private-key.json file missing, but doesn't check if the credentials file is valid (for instance, if it had been revoked or is for the wrong project). We can do a pre-flight check on before doing the api server deployment by using the keyfile to create a signed url and check if it works

wnojopra:
Fixes #375 

This test checks that your private-key.json file is valid by generating a signed url to an object in the project's bucket, and checks that we can successfully GET the signed url. 

I ran the script with a valid key file, and it passes. I verified that running the script with an invalid key file or a key file from a different project will give an error.

### DIFF
--- a/deploy/deploy-api.sh
+++ b/deploy/deploy-api.sh
@@ -11,6 +11,27 @@
 set -o errexit
 set -o nounset
 
+function check_private_key {
+  echo "Checking for valid private key by generating a signed URL on a temp GCS file"
+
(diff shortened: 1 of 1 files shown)


melissachang (approved):
Nice

PR #378, opened 2020-08-19
SQL query is now returned by Facets GET API. EXTRA_FACET_INFO is removed from app config.

### LINKED ISSUE #377: SQL generated is incorrect sometimes for extra facets
AMP PD admin reported: "I've noticed that the cohort query is saved incorrectly for filtering criteria entered in cards that are not shown by default, but added from a search. ... This appears to be random, as I've performed the exact same actions, clicks and keystrokes and received different results."

I believe the problem is only visible for time series facets in Extra Facets (that is, the facets that are not visible by default). The root cause is we are setting `EXTRA_FACET_INFO` in the `current_app.config`. `EXTRA_FACET_INFO` should be set per-session or per-request, but instead it is in a global dictionary that every user session touches. 

Reproducible steps:
0. Open two tabs, each open to https://time-series-data-explorer.appspot.com/
1. In tab 1, search for "Angina" (an extra facet), add the Angina facet, and select any of the values in the angina facet. 
2. In tab 2, search for "BMI" (a different extra facet), add the BMI facet, and select any of the values in the BMI facet.
3. Go back to tab 1, and save the cohort. After saving the cohort, you'll notice the SQL is incorrect. You'll get: 
```
SELECT DISTINCT t1.RANDID FROM ((SELECT RANDID FROM `verily-public-data.framingham_heart_study_teaching.framingham_heart_study_teaching.ANGINA` WHERE (1 = 1))) t1
```
What I expected to get was
```
SELECT DISTINCT t1.RANDID FROM ((SELECT RANDID FROM `verily-public-data.framingham_heart_study_teaching.framingham_heart_study_teaching` WHERE (ANGINA = 1 AND PERIOD = 1))) t1
```

I believe in step 2, we end up overwriting the EXTRA_FACET_INFO dictionary. In step 3, Angina no longer exists in the EXTRA_FACET_INFO dictionary, which is what we use to determine if the facet is a time series or not. Since it is not in the dictionary, we incorrectly determine that it is not a time series, and we produce the wrong SQL. 

To fix, I believe we should only be reading and writing EXTRA_FACET_INFO to dictionary scoped to the request.

wnojopra:
Fixes #377 

Summary of changes:

1) Code that constructs the SQL query was moved out of export_url_controller.py and into its own query_util.py file.
2) The facets_get API in facets_controller.py was updated to call get_sql_query and returns the sql_query in its response.
3) Client side javascript was updated to store the sql_query in app state. Currently, its only use is to be sent back to the server when calling ExportAPIPost. It will be used more in the near future as we implement new UI features that require this sql query.
4) export_url_controller.py no longer calls get_sql_query, instead the sql_query is passed from the client. 

You may notice auto-generated code changes in ui/src/api/src/model/*.js and api/data_explorer/models/*.py. This is due to my changes in api/app.yaml which introduces the sql_query parameter into the FacetsResponse and exportUrlRequest.


melissachang (commented):
Nice sleuthing. Making sql_query state in react SG. But if you're going to make it state, does sql_query state update when user clicks around the UI? (It should.)

melissachang (commented):
Why put sql query in React state, instead of just making the API server sql query computation request-scoped? Latter seems like a better solution. You have everything you need to know to generate sql query in url.

wnojopra:
> Nice sleuthing. Making sql_query state in react SG. But if you're going to make it state, does sql_query state update when user clicks around the UI? (It should.)

Yes it does. Clicking around the UI makes a GET facet request. We set the sql_query state in the callback. 

> Why put sql query in React state, instead of just making the API server sql query computation request-scoped? Latter seems like a better solution. You have everything you need to know to generate sql query in url.

That was also my initial hunch, so good question. Before this PR, The facets_controller cached extra facet data from Elasticsearch into EXTRA_FACET_INFO. The sql query generation code looked at  [EXTRA_FACET_INFO](https://github.com/DataBiosphere/data-explorer/blob/c6d25287631286ef9de0de29fe2643cb4e411fb0/api/data_explorer/controllers/export_url_controller.py#L214) to determine if a field is time series, and also to get the bucket intervals. Now that we realized this EXTRA_FACET_INFO isn't scoped per request/user-session, I have decided to remove it, and the sql generation code is now missing this information. We could instead update the sql generation code to fetch this data from elasticsearch, but it currently has zero dependencies on elasticsearch and its nice to keep it that way.

melissachang (approved):
Nice! Worked great with local demo, the SQL queries worked in BQ. Didn't look at code closely, I'll let Matt do that. :)

